PR #200, opened 2018-04-23
Use the file path if necessary when computing React display names

alangpierce:
Following the same semantics as the babel plugin, if there's a default-exported
react class, then we infer the name via the filename. I intentionally didn't use
the node `path` library to avoid a dependency on node.

### DIFF
--- a/src/transformers/ReactDisplayNameTransformer.ts
+++ b/src/transformers/ReactDisplayNameTransformer.ts
@@ -1,6 +1,7 @@
 import {ContextualKeyword, IdentifierRole} from "../../sucrase-babylon/tokenizer";
 import {TokenType as tt} from "../../sucrase-babylon/tokenizer/types";
 import ImportProcessor from "../ImportProcessor";
+import {Options} from "../index";
 import TokenProcessor from "../TokenProcessor";
 import RootTransformer from "./RootTransformer";
 import Transformer from "./Transformer";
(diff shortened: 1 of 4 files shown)


StreetStrider:
I don't get rationale for this. Is this for some React-specific debug tools? Looks like one, but I'm not an React expert to know for sure.

alangpierce:
@StreetStrider Yep! Sucrase's `ReactDisplayNameTransformer` mimics [babel-plugin-transform-react-display-name](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-react-display-name/src/index.js), which attaches a display name to `createReactClass` when possible. (This is for old-style React components; newer ones usually have the name explicitly attached as the class or function name.) The display name is mostly used for debugging tools like React devtools, but is also useful in tests and available at runtime.

StreetStrider:
@alangpierce, understood that, thank you.
I've found in discussions that there's a `production` flag for Sucrase and it turns off this meta-info. I personally just want to eliminate filename from browser bundle (at any mode), so `production` is good for me.

Maybe it's a good idea to put info related to production flag as a section in readme. I think some people may follow this question. I found some info in website/readme, but I think that relates to another tool's production mode, not Sucrase's one.

alangpierce:
@StreetStrider the `production` flag actually only affects the `__self` and `__source` keys when transforming JSX. I think it's typical to include display names in production (even ones inferred from the filename, though those are more obscure), though I guess with babel it's all pretty configurable. I'll file an issue around better documentation for that flag.

PR #210, opened 2018-05-20
Add support for custom JSX pragmas

### LINKED ISSUE #207: Suggestion: drop dependency on React.createElement
Not sure if you're aware, but JSX is used for other projects too, such as virtual-dom, snabbdom, surplus and others. It would definitely be worthwhile making the element factory function configurable, rather than hardcoding it to React.createElement.

alangpierce:
Fixes #207

As with Babel, we assume the pragma is a dot-separated list of identifiers (or
just an identifier), which we parse so that we can replace the base name if
necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,14 @@ The following proposed JS features are built-in and always transformed:
 * [Optional catch binding](https://github.com/tc39/proposal-optional-catch-binding):
   `try { doThing(); } catch { }`.
 
-When using the `import` transform, there are some options to enable legacy
-CommonJS interop approaches:
+### JSX Options
+Like Babel, Sucrase compiles JSX to React functions by default, but can be
+configured for any JSX use case.
+* **jsxPragma**: Element creation function, defaults to `React.createElement`.
+* **jsxFragmentPragma**: Fragment component, defaults to `React.Fragment`.
+
+### Legacy CommonJS interop
+Two legacy modes can be used with the `import` tranform:
 * **enableLegacyTypeScriptModuleInterop**: Use the default TypeScript approach
   to CommonJS interop instead of assuming that TypeScript's `--esModuleInterop`
   flag is enabled. For example, if a CJS module exports a function, legacy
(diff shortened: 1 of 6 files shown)


StreetStrider:
Hello. Is [pragma comment](https://gitlab.com/Rich-Harris/buble/issues/195) supported?

alangpierce:
@StreetStrider No, not at the moment. I just filed https://github.com/alangpierce/sucrase/issues/479 to track the issue. What's your use case, out of curiosity? Do you have a codebase where different files need different JSX behavior?

StreetStrider:
@alangpierce I haven't got specific use-case right now, I'm learning things here. I just previously knew that jsx pragma is a valid way to setup a vdom constructor. I'm not sure about real user base, but if there is such a way, this pragma will occur in real world JSX.

I think my future use-case would be so: explicitly use jsx pragma comment in every file and explicitly import corresponding vdom constructor. It is explicit and does not depend on external config of any kind. However, this is a still thing to investigate for me.

alangpierce:
@StreetStrider Makes sense. Yeah, JSX is a bit awkward in that it started out as a React thing and was made more general, so you need custom configuration (or pragma comments) to configure the transpile output. Seems like a roadblock making it even less likely that it'll eventually get native browser support.

I think in practice it tends to happen for projects using two different UI frameworks, e.g. a project that's in a transition between React and Vue. But it's true that pragma comments have the nice aspect of making code more universal and less reliant on external config.

StreetStrider:
@alangpierce I will be notified here. However, if this feature is too complex for Sucrase, I'm OK if it will not be implemented.
I'm attracted by simplicity of Sucrase's usage. Right now I'm using `tsc + noEmit` to typecheck + Sucrase to cut TypeScript syntax (previously, same for Flow). So, I'm always produce JS with Rollup + plugin infrastructure; I find that is a robust approach (I'm not using tsc for bundling, or Babel, or anything else — it's always Rollup, so I can freely reuse configs, plugins and knowledge).

Hopefully, JSX can be resolved in Sucrase, but if not, I will investigate `rollup-plugin-jsx` as a second step. I'm pretty sure it supports all the features, but involves Babel internally as a drawback.

PR #219, opened 2018-05-29
Expand copyright to include babylon/acorn authors, add AUTHORS file

alangpierce:


kittens:
You just need to put the `LICENSE` and `AUTHORS` file into the `sucrase-babylon` folder and you're fine.

alangpierce:
Ah, I see. I was planning on integrating it into the rest of the code pretty soon anyway; it's just tech debt that it's still its own top-level folder, rather than any significant attempt at keeping it a coherent fork or independent module. Does the top-level license approach make sense to you given that interpretation? Happy to add LICENSE and AUTHORS in the subdirectory regardless if you think that's better.

PR #221, opened 2018-05-29
add rollup-plugin-sucrase

Rich-Harris:
hey, nice work! I whipped up a Rollup plugin — https://github.com/rollup/rollup-plugin-sucrase — and it's already [speeding up my builds](https://github.com/sveltejs/svelte/pull/1511) by a decent factor. I'm also impressed by how much leaner the output is. The only drawback I can see currently is that sourcemaps don't seem to be emitted (unless I've misunderstood?); look forward to being able to do that.

thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,8 +114,7 @@ const compiledCode = transform(code, {transforms: ["typescript", "imports"]});
 
 There are also integrations for
 [Webpack](https://github.com/alangpierce/sucrase/tree/master/integrations/webpack-loader),
-[Gulp](https://github.com/alangpierce/sucrase/tree/master/integrations/gulp-plugin),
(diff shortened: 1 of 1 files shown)


alangpierce:
🎉 thanks! If you want, you're welcome to publish under `@sucrase/rollup-plugin`, but having it under the rollup project certainly seems quite reasonable.

Yeah, the source map story is a little sad still, and something I've been meaning to look into. The current state is that there aren't source maps, but the output intentionally preserves line numbers. I'll file an issue for follow-up discussion/tracking.

StreetStrider:
Thanks for all your work, both of you. :+1: Tried to replace `rollup-plugin-flow` with this one, works like a charm.

PR #236, opened 2018-06-07
Fix 'export * from' with enableLegacyTypeScriptModuleInterop

aleclarson:


alangpierce:
Thanks! Can you add a test? I think typescript-test.ts is the best place.

aleclarson:
There are no tests for the `enableLegacyTypeScriptModuleInterop` option yet, so I didn't bother. I'm not knowledgable on the potential edge cases. I just noticed this while working on another upcoming pull request. Sorry!

alangpierce:
I think you should be able to either add a param to `assertTypeScriptResult` or make a similar assertion function. Just a simple case should be fine for this sort of thing, but I would prefer tests for every change, and actually find it to be a really nice aspect of working on Sucrase.

And yeah, I guess I was a bit lazy when adding `enableLegacyTypeScriptModuleInterop` and updated tests for the new output but didn't add tests for both cases. See https://github.com/alangpierce/sucrase/pull/202/files#diff-763ef3d30486c4fcadd81744662399ff for the old tests.

Issue #223, opened 2018-05-29
Generate source maps


alangpierce:
Mentioned by @Rich-Harris in #221.

Currently, Sucrase preserves line numbers, which gets many of the benefits of source maps (like reliable stack traces when using `sucrase/register`), but ideally it would emit real source maps. If we just care about lines, I think it would be a trivial source map that maps each line to itself. Mapping tokens within each line (which I think is useful for debuggers) should also be doable, but would need to be more integrated into the code generator.

aleclarson:
Have you started on this yet? This is my most desired feature. :)

alangpierce:
Nope, not yet. I'm hoping to get to it this weekend. Happy to collaborate if you want to take any of it on or do some research. It'll probably require a breaking change (currently Sucrase just returns a string, and I guess it'll need to return an object with both), so that's another thing I'll need to figure out.

aleclarson:
I might be able to help on the import/export side of things, but I have no time (or desperate need) for TS and Flow. Also, you should create a `sourcemaps` branch once you get started (and add me as a contributor) so I don't need to ask you about progress. Cheers!

alangpierce:
@aleclarson do you know what the best way is to test that source maps are working right? I actually don't have a great sense of how tools use source mappings within lines. Maybe it's needed for hover-to-view-value when debugging in Chrome DevTools (or WebStorm or VSCode), or if declared names get transformed, it uses the mapping to allow referencing the source code name from the console?

Also, my current plan is to have two modes (like magic-string has): one that only does line mappings and one that maps everything. Line mappings are a little pointless since the lines all correspond to the original source code, but I could see it being nice to have something in the source map format that you can feed into other tools. Do you think you'd use the full mode?

alangpierce:
Oh, and hopefully this can be done at a lower level than any particular transform and automatically work for all of them at once. I should be able to get started on it pretty soon (the 3.0 release unblocked this since now we can return more than just the code), but I'm a bit new to the details of the sourcemap format, so I'll need to do some research first.

aleclarson:
> do you know what the best way is to test that source maps are working right?

I'm not aware of any automated source map testing tools. 😢 I think unit-testing the decoded mappings and manual debugger testing are the only way (but I could be mistaken).

[...]

PR #257, opened 2018-06-17
Add basic support for source maps

alangpierce:
Progress toward #223

This is a first pass that's meant to be simple, flexible, and fairly
unopinionated. Notably, it does NOT yet compute mappings within each line, and
instead just maps each line to the corresponding line in the source. I'll
implement smarter mappings in the future, possibly as an option if it ends up
hurting performance.

A quick benchmark shows that enabling source maps increases total running time
by about 25%, even with this simple approach. Hopefully that can be optimized in
the future.

Another possible to-do item is for Sucrase to accept a source map which it then
modifies. However, this can hopefully be handled by other tools that compose
source maps.

There are also various tweaks that may make this more convenient, depending on
use case.
* Optionally include the original source code in `sourcesContent` in the source
  map.
* Optionally append a sourceMappingURL comment to the end of the source code.
* Optionally return the source map as a string instead of an object.
* Better control over the sourceRoot and file paths specified in the source map.

All of these use cases can be satisfied with just one or two lines of code after
running Sucrase, though, so none seem super critical.

Some docs of other tools emitting source maps, which I used for inspiration:
https://babeljs.io/docs/en/babel-core
https://www.typescriptlang.org/docs/handbook/compiler-options.html

### DIFF
--- a/src/computeSourceMap.ts
+++ b/src/computeSourceMap.ts
@@ -1,0 +1,22 @@
+import sourceMap, {RawSourceMap} from "source-map";
+
+/**
+ * Generate a simple source map indicating that each line maps directly to the original line.
+ */
+export default function computeSourceMap(code: string, filePath: string): RawSourceMap {
+  const mapGenerator = new sourceMap.SourceMapGenerator({file: filePath});
+  let numLines = 1;
+  for (let i = 0; i < code.length; i++) {
+    if (code[i] === "\n") {
+      numLines++;
+    }
+  }
+  for (let line = 1; line <= numLines; line++) {
+    mapGenerator.addMapping({
+      source: filePath,
+      generated: {line, column: 0},
+      original: {line, column: 0},
+    });
+  }
+  return mapGenerator.toJSON();
+}
--- a/test/source-maps-test.ts
+++ b/test/source-maps-test.ts
@@ -1,0 +1,22 @@
+import * as assert from "assert";
+import {transform} from "../src";
+
+describe("source maps", () => {
+  it("generates a simple line-based source map", () => {
+    const result = transform(
+      `\
+      import a from "./a";
+      const x: number = 1;
+      console.log(x + 1);
+    `,
+      {transforms: ["imports", "typescript"], computeSourceMap: true, filePath: "test.ts"},
+    );
+    assert.deepEqual(result.sourceMap, {
+      version: 3,
+      sources: ["test.ts"],
+      names: [],
+      mappings: "AAAA;AACA;AACA;AACA",
+      file: "test.ts",
+    });
+  });
+});
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ self-build
 dist
 example-runner/example-repos
 integrations/gulp-plugin/dist
+.nyc_output
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "eslint-plugin-prettier": "^2.4.0",
     "eslint-plugin-typescript": "^0.12.0",
     "mocha": "^3.5.3",
+    "nyc": "^12.0.2",
     "prettier": "^1.12.1",
     "sucrase": "^2.2.0",
     "tslint": "^5.9.1",
@@ -70,7 +71,8 @@
     "commander": "^2.12.2",
     "lines-and-columns": "^1.1.6",
     "mz": "^2.7.0",
-    "pirates": "^3.0.2"
+    "pirates": "^3.0.2",
+    "source-map": "^0.7.3"
   },
   "engines": {
     "node": ">=8"
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,6 @@
+import {RawSourceMap} from "source-map";
 import CJSImportProcessor from "./CJSImportProcessor";
+import computeSourceMap from "./computeSourceMap";
 import identifyShadowedGlobals from "./identifyShadowedGlobals";
 import NameManager from "./NameManager";
 import {parse} from "./parser";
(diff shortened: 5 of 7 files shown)


alangpierce:
Still needs tests.

@aleclarson let me know if you have thoughts on the details here!

aleclarson:
> Another possible to-do item is for Sucrase to accept a source map which it then
modifies. However, this can hopefully be handled by other tools that compose
source maps.

[Rich-Harris/sorcery](https://github.com/Rich-Harris/sorcery) was made for that, and [mozilla/source-map](https://github.com/mozilla/source-map) has [`applySourceMap`](https://github.com/mozilla/source-map/#sourcemapgeneratorprototypeapplysourcemapsourcemapconsumer-sourcefile-sourcemappath), so I don't think it's worth it.

Side note: My fork of `sorcery` [avoids lossy merging](https://github.com/aleclarson/sorcery/issues/1) of source maps, which is a common issue. :)

> Optionally include the original source code in sourcesContent in the source
map.

I think it's considered "good practice" to do this by default. Babel, CoffeeScript, and Bublé all do it this way. Babel and CoffeeScript don't even provide an opt-out, but I would prefer an `includeContent` option that disables this feature when `false`.

> Optionally append a sourceMappingURL comment to the end of the source code.

I'm in favor of a `sourceMaps` option that behaves like below:

The `sourceMaps` option defaults to falsy, which means no sourcemap is generated. Setting `sourceMaps` to `true` or `"both"` will generate a `SourceMap` object and return it as the `map` property of the result object. Setting `sourceMaps` to `"inline"` or `"both"` will append a `//# sourceMappingURL` comment to the generated code. When `sourceMaps` equals `"inline"` or falsy, the `process` function returns a string (the generated code) instead of an object.

That is how [nebu](https://github.com/aleclarson/nebu) does it. ([see here](https://github.com/aleclarson/nebu/blob/master/src/index.coffee#L45-L65))

> Optionally return the source map as a string instead of an object.

The complexity isn't worth it, IMO. It's super easy to `JSON.stringify` the object.

> Better control over the sourceRoot and file paths specified in the source map.

The `sourceRoot` property is pointless when there's only a single source path. I suggest having `source` and `file` options [like `magic-string` does](https://github.com/Rich-Harris/magic-string/blob/master/src/MagicString.js#L164-L165).

src/computeSourceMap.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export default function computeSourceMap(code: string, filePath: string): RawSourceMap {
+  const mapGenerator = new sourceMap.SourceMapGenerator({file: filePath});
+  let numLines = 1;
+  for (let i = 0; i < code.length; i++) {
+    if (code[i] === "\n") {
+      numLines++;
+    }
+  }
+  for (let line = 1; line <= numLines; line++) {
```
aleclarson: You only need to call `addMapping` for changed lines, I think.

test/source-maps-test.ts
```diff
@@ -0,0 +1,22 @@
...
+      console.log(x + 1);
+    `,
+      {transforms: ["imports", "typescript"], computeSourceMap: true, filePath: "test.ts"},
+    );
+    assert.deepEqual(result.sourceMap, {
+      version: 3,
+      sources: ["test.ts"],
+      names: [],
+      mappings: "AAAA;AACA;AACA;AACA",
+      file: "test.ts",
```
aleclarson: The `file` property is supposed to be the *generated* filename.

alangpierce: Since Sucrase just returns a string rather than writing anything to the filesystem, I don't know what file name it'll be written as (if any). And if it's an intermediate step like in webpack, maybe it wouldn't have an individual output filename at all? I suppose I could just replace the extension with `.js`, do you think that would be better? One thought is the caller can always swap this out with whatever filename they intend to persist the result as, but I suppose an option given to Sucrase can also work.

aleclarson: I suggest having `source` and `file` options [like `magic-string` does](https://github.com/Rich-Harris/magic-string/blob/master/src/MagicString.js#L164-L165).

src/computeSourceMap.ts
alangpierce: I tried debugging in WebStorm with no mappings at all and it didn't work, so maybe in at least some cases you need at least one line per mapping? Certainly in some configurations/situations, Sucrase won't change the code at all, and I still want the resulting file to be debuggable.

Also, at least for this simple approach (which may or may not be reasonable in the long run), it's a lot easier to just do one mapping per line rather than checking if the line changed.

aleclarson: You may need a mapping for the `{line: 0, column: 0}` for my suggestion to work. Otherwise, the debugger will assume the code before the first mapping is untraceable. Can you try that?

For the interim solution this PR provides, it's nothing to worry about. But it's worth avoiding (if possible) when you get around to implementing column mappings.

alangpierce: I just played around with it a bit, and it looks to me like I actually do need to map every line, at least for it to work in WebStorm. With an untransformed code snippet that's 4 lines long, if I just map the first and third line and set breakpoints on every line, it only stops on the lines I explicitly mapped, and grays out the breakpoints on unmapped lines to indicate that they won't do anything:
screen shot 2018-06-17 at 12 59 38 pm

This might just be a bug in WebStorm, but obviously I want to do whatever makes tools work. OTOH, this strategy of mapping just the start of each line seems to get everything I want in WebStorm, even variable hovering on unmapped lines (which presumably just looks at the identifier name under the mouse cursor and tries to resolve it).

So for now, I'll leave this as-is.

aleclarson: Strange. You might want to notify the WebStorm team. Have you tested that same example in Chrome, too?

alangpierce: Just tested it in Chrome now (chrome devtools for node, not running JS on a webpage), and looks like Chrome has the same behavior:

screen shot 2018-06-17 at 8 25 10 pm

The grayed out lines means I can't set a breakpoint, and stepping line-by-line skips those lines. So looks like tools expect at least one mapping per line? Should be pretty easy to include on my end, I think.

PR #268, opened 2018-06-20
JSX props values can be elements too

### LINKED ISSUE #267: JSX props can be elements too
The following fails:

```js
const x = <a b=<c/> d="e"/>;
```

https://sucrase.io/#compareWithTypeScript=true&code=%0Aconst%20x%20%3D%20%3Ca%20b%3D%3Cc%2F%3E%20d%3D%22e%22%2F%3E%3B

See JSXAttributeValue at https://facebook.github.io/jsx/

arv:
Fixes #267

The value of a JSX attribute/prop can be a JSX element or a fragment.

See JSXAttributeValue at https://facebook.github.io/jsx/

### DIFF
--- a/src/parser/plugins/jsx/index.ts
+++ b/src/parser/plugins/jsx/index.ts
@@ -267,6 +267,9 @@ function nextJSXTagToken(): void {
       case charCodes.greaterThan:
         finishToken(tt.jsxTagEnd);
         break;
+      case charCodes.lessThan:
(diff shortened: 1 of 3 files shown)


arv:
Let me know if you would prefer 2 PRs instead of 2 commits. They are independent except for the test.

alangpierce:
Thank you! Looks good to me! (For larger things, 2 PRs would be nice, but here it's fine to combine them.)

PR #270, opened 2018-06-23
Add production option and use it for JSX

arv:
When production is true (default is false) we do not include debug
info (such as `__self` and `__source`) in the output of the jsx
transformer.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -48,8 +48,7 @@ export interface Options {
    */
   filePath?: string;
   /**
-   * If specified, we generate a few less things (for example the jsx
-   * transformer does not add debug props).
+   * If specified, omit any development-specific code in the output.
    */
   production?: boolean;
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -47,6 +47,11 @@ export interface Options {
    * File path to use in error messages, React display names, and source maps.
    */
   filePath?: string;
+  /**
+   * If specified, we generate a few less things (for example the jsx
+   * transformer does not add debug props).
+   */
+  production?: boolean;
 }
 
 export interface TransformResult {
--- a/src/transformers/JSXTransformer.ts
+++ b/src/transformers/JSXTransformer.ts
@@ -83,14 +83,28 @@ export default class JSXTransformer extends Transformer {
 
   processProps(firstTokenStart: number): void {
     const lineNumber = this.getLineNumberForIndex(firstTokenStart);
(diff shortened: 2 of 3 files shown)


src/index.ts
```diff
@@ -47,6 +47,11 @@ export interface Options {
    * File path to use in error messages, React display names, and source maps.
    */
   filePath?: string;
+  /**
+   * If specified, we generate a few less things (for example the jsx
```
arv: Suggestions wanted.

src/transformers/JSXTransformer.ts
```diff
@@ -83,14 +83,28 @@ export default class JSXTransformer extends Transformer {
...
+      if (devProps) {
+        this.tokens.appendCode(`, {${devProps}}`);
+      } else {
+        this.tokens.appendCode(`, null`);
+      }
       return;
     }
     this.tokens.appendCode(`, {`);
+    let firstProperty = true;
+    const appendComma = () => {
```
arv: Any ideas how to make this cleaner?

arv:
I guess I should update README too?

src/transformers/JSXTransformer.ts
alangpierce: Is the `appendComma` change just there to avoid the trailing comma? It's fine to assume that any compilation target supports trailing object commas, so I'd prefer to just keep the trailing comma in the output for simplicity.

The original code before I added `devProps` was to just always have a trailing comma: https://github.com/alangpierce/sucrase/pull/176/files#diff-17d5ba94e1a4d345c1187a5cb0b54ad7R96

Also, the closure approach here is unlikely to work with AssemblyScript (which currently doesn't work on all of Sucrase, but I'd like to move in a direction where it eventually can), so simpler control flow would be better here.

src/index.ts
alangpierce: How about this: "If specified, omit any development-specific code in the output."

It can be changed later if more use cases come up that don't exactly fit that.

test/jsx-test.ts
```diff
@@ -413,4 +419,66 @@ describe("transform JSX", () => {
       {extraTransforms: ["imports"], jsxPragma: "h", jsxFragmentPragma: "Fragment"},
     );
   });
+
+  describe("with production true", () => {
```
alangpierce: Thanks for the tests!

arv:
PTAL

PR #285, opened 2018-06-29
Don't emit semicolons in class bodies

### LINKED ISSUE #280: Semicolons in class bodies seem to cause problems with webpack
Example:
```js
class A {
  b = 1;
}
```
becomes this:
```js
"use strict";class A {constructor() { this.b = 1; }
  ;
}
```
[Playground link](https://sucrase.io/#code=class%20A%20%7B%0A%20%20b%20%3D%201%3B%0A%7D)

This behavior changed in Sucrase 3.2.0 ( #271 ), and it caused Sucrase 3.2.1 to be compiled in a way that broke the webpack build in the sucrase.io website. For some reason, it causes webpack 3.5.1 to silently stop about 15 seconds in with no error message. For now, it seems best to change the behavior to avoid ever emitting a lone semicolon.

alangpierce:
Fixes #280.

It's unclear what the real problem is, but this broke the website build, and is
probably cleaner to remove anyway.

### DIFF
--- a/src/util/getClassInfo.ts
+++ b/src/util/getClassInfo.ts
@@ -51,6 +51,7 @@ export default function getClassInfo(
     if (tokens.matchesContextual(ContextualKeyword._constructor)) {
       ({constructorInitializers, constructorInsertPos} = processConstructor(tokens));
     } else if (tokens.matches1(tt.semi)) {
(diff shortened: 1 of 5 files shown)


alangpierce:
Build passed but for some reason isn't showing up in the PR: https://travis-ci.org/alangpierce/sucrase/builds/398046506

PR #304, opened 2018-07-20
Add disableLegacyClassFieldSupport option

### LINKED ISSUE #302: Eventually is almost here: Class Fields
https://github.com/alangpierce/sucrase/blob/15debf701596bd2a1d7564d8f1e147da357dfb56/src/transformers/RootTransformer.ts#L188-L191

Could we make those optional? There is partial implementation in V8 behind a flag.

Thanks 👍

SMotaal:
This adds the ability to opt out of class fields transformations. I imagine this to eventually be promoted to enableLegacy.

Closes #302

### DIFF
--- a/src/transformers/RootTransformer.ts
+++ b/src/transformers/RootTransformer.ts
@@ -24,7 +24,7 @@ export default class RootTransformer {
     sucraseContext: SucraseContext,
     transforms: Array<Transform>,
     enableLegacyBabel5ModuleInterop: boolean,
-    options: Options,
+    private options: Options,
   ) {
     this.nameManager = sucraseContext.nameManager;
     const {tokenProcessor, importProcessor} = sucraseContext;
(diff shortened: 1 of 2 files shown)


SMotaal:
@alangpierce Not sure why it stalls... All tests ran smoothly on my end. Do I need to run some prep script before commits intended to CI?

alangpierce:
Thanks! I'll take a look when I get a chance. Not sure about the build, I haven't seen that error before. I'll retry and hopefully that'll fix it.

src/transformers/RootTransformer.ts
```diff
@@ -210,25 +210,31 @@ export default class RootTransformer {
...
-        }
-        fieldIndex++;
-      } else if (this.tokens.currentIndex() === constructorInsertPos) {
-        this.tokens.copyToken();
-        if (initializerStatements.length > 0) {
-          this.tokens.appendCode(`;${initializerStatements.join(";")};`);
-        }
-        this.processToken();
-      } else {
+    if (this.options.disableLegacyClassFieldSupport) {
```
alangpierce: There are a few details here to think through:
* I was about to say "we don't need to call `processClass` at all", but it actually still is necessary for TypeScript constructor initializers.
* You will still need to emit `initializerStatements`, at least in the TypeScript case, since constructor initializers aren't part of JS.
* The `if` statement above this handles the case where the class didn't have any constructor at all, and in that case, it inserts a constructor. You'll need to disable that case as well.
* Note that `getClassInfo` is where a lot of the complexity lives. I was hoping that we'd be able to avoid calling it, but TypeScript constructor initializers will still need it. Or maybe we could make that specific transform much more focused if it's the only thing that needs to be done.

SMotaal: Perfect... I did not dive deep and wanted to wait for insights like this since it is my first contribution and I learned the hard lesson of not reinventing the wheel.

For my case, the code works only because I am not relying on any TypeScript features, my modules are already 99.9% mjs, but for most this is not the case.

Since this requires a specific node version and flags, I think testing beyond string validation will require a complicated infra.

Based on all your remarks, I am leaning towards a completely different take. I know that speed and simplicity are top a priorities, and that obviously means plugins are a no go (and I love that). 

So in my opinion, one idea to consider is creating a dedicated transform path `customTransform` which would follow the same template method as the normal transform with opt-in overloads hooks that are strategically thought out. I realize that this is not within the scope of sucrase in itself but it is within the scope of experimentation that makes it possible to make contributions across the board (sucrase  included).

What are your thoughts?

SMotaal: @alangpierce any feedback would be great

alangpierce: Hi, sorry for not getting back to this sooner!

> one idea to consider is creating a dedicated transform path customTransform which would follow the same template method as the normal transform with opt-in hooks that are strategically thought out

In my view, this is actually just a plugin system 😄. Or arguably it already exists with the `Transformer` abstract class, but keep in mind that `Transformer` is *not* actually flexible enough for what Sucrase needs, and there are various hacks that get different combinations to work outside of that system. We could try to make `Transformer` better, but I think there probably isn't a clean set of hooks that makes transforms nicely-composable. I think your approach here of plain `if` statements is actually the cleanest way forward, as long as we only implement features that have a strong justification (and I think yours does).

> I think testing beyond string validation will require a complicated infra.

Agreed! And, in fact, nearly all tests right now are just string validation. You should be able to add new tests at the bottom of sucrase-test.ts that use your new option without actually running the compiled code.

> I am not relying on any TypeScript features

Makes sense! Here's my current take on what we should do:

* With `disableLegacyClassFieldSupport` enabled and `typescript` disabled: make `processClass` use a much simpler code path that just processes tokens to the end of the class without doing anything special. I think you can just call `processBalancedCode` at the open-brace and it will do normal transformation up to the close-brace. Note that we still need the logic of skipping to the end of the class, since it's needed by `CJSImportTransformer.processExportClass`.
* With `disableLegacyClassFieldSupport` enabled and `typescript` enabled: throw an error saying that this combination is unsupported. I'll probably want to implement it eventually, but I think it's fine to have unsupported combinations of options and add support later.

SMotaal: I think ruling out `typescript` enabled really limits the scope of sucrase, at least in my case. The use case of authoring with Type Definitions (ie TS files) and being able to quickly strip away type annotations (ie `--experimental-modules` for now with `--loader imagine-sucrase-loader`) is going to become more relevant in the next months.

Do you think it is fair to say there is too much coupling. I would like to play around with the idea of a separate transformer all together, the Typeless transform. It's a type stripper, which can be extended to allow type-aware transformations at the same time. My thought is that certain transforms depend on type information. Any thoughts on this front?

alangpierce:
Closing since this is now addressed with the `disableESTransforms` flag.

PR #318, opened 2018-10-04
Add codecov

### LINKED ISSUE #258: Add code coverage in CI
Given that part of the goal of this project is to have a small amount of highly-reliable code, seems like having really good coverage and enforcing it is a good way to showcase that and keep the project honest. Current stats from running in WebStorm show 82% coverage, so seems like there could be some improvement. Seems like https://coveralls.io/ might be the best way to do that.

raylu:
Fixes #258

Add coverage via nyc/istanbul and report with codecov

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
         - yarn run-examples tslint apollo-client
     - script:
         - yarn build
-        - yarn test
+        - 'yarn test && yarn coverage'
         - yarn run-examples decaffeinate decaffeinate-parser coffee-lex
     - node_js: '8'
       script:
--- a/package.json
+++ b/package.json
@@ -24,8 +24,9 @@
     "prepublishOnly": "yarn clean && yarn build",
     "release": "sucrase-node script/release.ts",
     "run-examples": "sucrase-node example-runner/example-runner.ts",
-    "test": "yarn lint && yarn test-only",
-    "test-only": "mocha './test/**/*.ts'"
+    "test": "yarn lint && nyc yarn test-only",
+    "test-only": "mocha './test/**/*.ts'",
+    "coverage": "nyc report --reporter=text-lcov > coverage.lcov && codecov"
   },
   "repository": {
     "type": "git",
@@ -53,6 +54,7 @@
     "@types/mocha": "^2.2.43",
     "@types/mz": "^0.0.32",
     "@types/node": "^8.0.31",
+    "codecov": "^3.1.0",
     "eslint": "^4.13.1",
     "eslint-config-airbnb-base": "^12.1.0",
     "eslint-config-prettier": "^2.9.0",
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -30,7 +30,7 @@ matrix:
     - node_js: '8'
       script:
         - yarn build
-        - yarn test
+        - 'yarn test && yarn coverage'
```
alangpierce: Could you move this to the `yarn test` in the build above? The "normal" build config is node 10, and this build config is the one to make sure I didn't break node 8.

test/mocha.opts
```diff
@@ -1 +1,3 @@
+--compilers ts-node/register
```
alangpierce: `--compilers` is deprecated in mocha: https://github.com/mochajs/mocha/wiki/compilers-deprecation

You can just do `--require` like the other two.

package.json
```diff
@@ -62,7 +64,9 @@
     "mocha": "^3.5.3",
     "nyc": "^12.0.2",
```
alangpierce: Heh, I forgot I already had this installed.

package.json
```diff
@@ -76,5 +80,14 @@
   },
   "engines": {
     "node": ">=8"
+  },
+  "nyc": {
```
alangpierce: Can you move this to a .nycrc file? I generally prefer to avoid dev tool config in package.json (though I realize it's reasonably common).

test/mocha.opts
alangpierce: Did you try this without `ts-node/register`? Sucrase should generally work, and produces inline source maps (it actually always preserves line numbers, so source maps are trivial), but maybe it crashes nyc?

In any case, `ts-node/register` and `sucrase/register` both set up require hooks to compile typescript on the fly, so if we set up both hooks, one will just clobber the other, so we should pick one. I'd prefer for tests to be compiled with sucrase when run normally (I did some intentional work to switch to that), though if that's a pain, I'm fine switching back to ts-node.

.travis.yml
```diff
@@ -25,7 +25,7 @@ matrix:
         - yarn run-examples tslint apollo-client
     - script:
         - yarn build
-        - yarn test
+        - 'yarn test && yarn coverage'
```
alangpierce: Any reason not to put them on separate lines? I think the travis build will fail if you get a non-zero exit code from any of these. I guess maybe you could say they're semantically linked, in that running coverage without test doesn't make sense.

Also, why the quotes? I guess that's a yaml thing?

raylu: > I think the travis build will fail if you get a non-zero exit code from any of these.
the build will fail but

https://docs.travis-ci.com/user/customizing-the-build/#customizing-the-build-step
> When one of the build commands returns a non-zero exit code, the Travis CI build runs the subsequent commands as well, and accumulates the build result.

and I don't want coverage to run on a failing build

> Also, why the quotes? I guess that's a yaml thing?

I wasn't sure what would happen with the `&`s. want me to remove them?

alangpierce: Nah, it's fine, maybe a minor cleanup in the future, but iterating on travis stuff is sort of a pain anyway.

alangpierce:
@raylu Thanks!

PR #324, opened 2018-10-26
Allow super.method() calls in constructor

### LINKED ISSUE #323: super does not work in constructor
```js
class Y extends X {
    constructor() {
        super();
        super.m();
    }
};
```

https://sucrase.io/#selectedTransforms=jsx&code=%0Aclass%20Y%20extends%20X%20%7B%0A%20%20%20%20constructor()%20%7B%0A%20%20%20%20%20%20%20%20super()%3B%0A%20%20%20%20%20%20%20%20super.m()%3B%0A%20%20%20%20%7D%0A%7D%3B%0A

Same error happens with `super.x` and `super.x = 1`

arv:
The old code assumed that `super` always followed by a `(` which is not
the case.

Also adds tests for invalid (at runtime) code with multiple `super()`
and `super.m()` / `super.x` / `super.x = 1` before `super()`.

Fixes #323

### DIFF
--- a/src/util/getClassInfo.ts
+++ b/src/util/getClassInfo.ts
@@ -209,17 +209,21 @@ function processConstructor(
   let constructorInsertPos = tokens.currentIndex();
 
   // Advance through body looking for a super call.
+  let foundSuperCall = false;
   while (!tokens.matchesContextIdAndLabel(tt.braceR, constructorContextId)) {
-    if (tokens.matches1(tt._super)) {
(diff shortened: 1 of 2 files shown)


arv:
@alangpierce PTAL

I added support and tests for multiple super() and super.a access before super() even though all those cases would lead to runtime errors. Is it worth dealing with this or can we just assume that we do not get this kind of invalid input code?

PR #327, opened 2018-11-05
Restrict TS try/catch logic to cases where backtracking is needed

### LINKED ISSUE #326: Performance regression with recent Babel parser changes
Benchmark results before and now:

**Sucrase 3.4.2:**
```
Simulating 100 compilations of sample.tsx:
Sucrase: 479.140ms
TypeScript: 3641.253ms
Babel: 6293.548ms
```
**master:**
```
Simulating 100 compilations of sample.tsx:
Sucrase: 4350.796ms
TypeScript: 3691.154ms
Babel: 6388.539ms
```

From some profiling, looks like the culprit is probably an extra `tsTryParseAndCatch` call added to `tsParseSubscript`. I think the throw/catch exception pattern is really slow and is now happening commonly in the benchmark. It's also spending quadratic time because it scans the entire file to compute the line/column number to put in the thrown error. Hopefully this benchmark is an especially bad case and the actual perf regression is less than 10x on realistic code.

alangpierce:
Fixes #326

When porting some babel-parser changes, there was some code in TS parsing that
was unconditionally wrapping some parsing in a try/catch, but actually it only
cared about cases that started with `<`. By inverting the logic a little, we
avoid having to throw and catch an exception in the common case of parsing,
which was dramatically slowing down TypeScript parsing.

### DIFF
--- a/src/parser/plugins/typescript.ts
+++ b/src/parser/plugins/typescript.ts
@@ -1093,10 +1093,10 @@ export function tsParseSubscript(
     return;
   }
 
-  // There are number of things we are going to "maybe" parse, like type arguments on
-  // tagged template expressions. If any of them fail, walk it back and continue.
-  const success = tsTryParseAndCatch(() => {
-    if (match(tt.lessThan)) {
+  if (match(tt.lessThan)) {
+    // There are number of things we are going to "maybe" parse, like type arguments on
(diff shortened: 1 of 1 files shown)


alangpierce:
Looks like something is buggy about coverage. I confirmed that it isn't related to this PR and filed as #329

PR #337, opened 2018-11-18
fix: #322

### LINKED ISSUE #322: imported function call without leading semicolon
[Repro link](https://sucrase.io/#code=import%20%7Bfun%7D%20from%20'.%2Ffoo'%0Alet%20arr%20%3D%20%5B%5D%0Afun())

```ts
import {fun} from './foo'
let arr = []
fun()
```

..becomes..

```ts
"use strict";var _foo = require('./foo');
let arr = []
(0, _foo.fun)()
```

..which evaluates to `TypeError: [] is not a function`

aleclarson:
This is a work-in-progress. Currently, it spits out an error.

```
Tried to finish processing tokens before reaching the end.
```

Fixes #322

### DIFF
--- a/benchmark/sample/expression.ts
+++ b/benchmark/sample/expression.ts
@@ -633,7 +633,7 @@ function parseParenItem(): void {
 // argument to parseSubscripts to prevent it from consuming the
 // argument list.
 function parseNew(): void {
-  parseIdentifier();
+  expect(tt._new);
   if (eat(tt.dot)) {
     // new.target
     parseMetaProperty();
--- a/src/transformers/CJSImportTransformer.ts
+++ b/src/transformers/CJSImportTransformer.ts
@@ -174,22 +174,52 @@ export default class CJSImportTransformer extends Transformer {
     const replacement = this.importProcessor.getIdentifierReplacement(
       this.tokens.identifierNameForToken(token),
     );
-    if (replacement) {
-      this.tokens.replaceToken(replacement);
-
-      // Any number of closing parens is tolerated.
-      while (!this.tokens.isAtEnd() && this.tokens.currentToken().type === tt.parenR) {
-        this.rootTransformer.processToken();
+    if (!replacement) {
+      return false;
+    }
+    let openParenIndex = this.tokens.currentIndex();
+    while (++openParenIndex < this.tokens.tokens.length) {
+      const token = this.tokens.tokens[openParenIndex];
+      if (token.type === tt.parenL) {
+        // Avoid treating imported functions as methods of their `exports` object
+        // by using `(0, f)` when the identifier is in a paren expression. Else
+        // use `Function.prototype.call` when the identifier is a guaranteed
+        // function call. When using `call`, pass undefined as the context.
+        switch (this.tokens.tokenAtRelativeIndex(1).type) {
+          case tt.parenL: {
+            // We can use `(0, f)` when the previous token is an open brace,
+            // semicolon, or equal operator. This saves space for function calls
+            // with arguments, because then `void 0` is omitted.
+            const prevToken = this.tokens.tokenAtRelativeIndex(-1);
+            if (
+              prevToken.type !== tt.braceL &&
+              prevToken.type !== tt.semi &&
+              prevToken.type !== tt.eq
+            ) {
+              // NOTE: This line inserts a new parenthesis, which must be accounted
+              // for by recursively calling the `processBalancedCode` method of the
+              // root transformer.
+              const hasArgs = this.tokens.tokenAtRelativeIndex(2).type !== tt.parenR;
+              this.tokens.replaceToken(`${replacement}.call(` + (hasArgs ? "void 0, " : ""));
+              this.tokens.removeToken(); // Remove the old paren.
+              this.rootTransformer.processBalancedCode(0, 1);
+              return true;
+            }
+          }
+          case tt.parenR:
+            // See here: http://2ality.com/2015/12/references.html
+            this.tokens.replaceToken(`(0, ${replacement})`);
+            return true;
+        }
+        break;
       }
-
-      // Avoid treating imported functions as methods of their `exports` object by using
-      // the `call` method with a falsy context (which is coerced into the global context).
-      if (!this.tokens.isAtEnd() && this.tokens.currentToken().type === tt.parenL) {
-        this.tokens.replaceToken(`.call(0, `);
-        return true;
+      // Tolerate any number of closing parens.
+      if (token.type !== tt.parenR) {
+        break;
       }
     }
-    return false;
+    this.tokens.replaceToken(replacement);
+    return true;
   }
 
   processObjectShorthand(): boolean {
--- a/src/transformers/CJSImportTransformer.ts
+++ b/src/transformers/CJSImportTransformer.ts
@@ -177,46 +177,32 @@ export default class CJSImportTransformer extends Transformer {
     if (!replacement) {
       return false;
     }
-    let openParenIndex = this.tokens.currentIndex();
-    while (++openParenIndex < this.tokens.tokens.length) {
-      const token = this.tokens.tokens[openParenIndex];
-      if (token.type === tt.parenL) {
-        // Avoid treating imported functions as methods of their `exports` object
-        // by using `(0, f)` when the identifier is in a paren expression. Else
-        // use `Function.prototype.call` when the identifier is a guaranteed
-        // function call. When using `call`, pass undefined as the context.
-        switch (this.tokens.tokenAtRelativeIndex(1).type) {
-          case tt.parenL: {
-            // We can use `(0, f)` when the previous token is an open brace,
-            // semicolon, or equal operator. This saves space for function calls
-            // with arguments, because then `void 0` is omitted.
-            const prevToken = this.tokens.tokenAtRelativeIndex(-1);
-            if (
-              prevToken.type !== tt.braceL &&
-              prevToken.type !== tt.semi &&
-              prevToken.type !== tt.eq
-            ) {
-              // NOTE: This line inserts a new parenthesis, which must be accounted
-              // for by recursively calling the `processBalancedCode` method of the
-              // root transformer.
-              const hasArgs = this.tokens.tokenAtRelativeIndex(2).type !== tt.parenR;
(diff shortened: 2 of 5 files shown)


src/transformers/CJSImportTransformer.ts
```diff
@@ -174,27 +174,22 @@ export default class CJSImportTransformer extends Transformer {
...
+
+      // Any number of closing parens is tolerated.
+      while (!this.tokens.isAtEnd() && this.tokens.currentToken().type === tt.parenR) {
+        this.rootTransformer.processToken();
+      }
+
+      // Avoid treating imported functions as methods of their `exports` object by using
+      // the `call` method with a falsy context (which is coerced into the global context).
+      if (!this.tokens.isAtEnd() && this.tokens.currentToken().type === tt.parenL) {
+        this.tokens.replaceToken(`.call(0, `);
```
alangpierce: This should have `void 0` or `undefined` as the arg, not just `0`.

alangpierce: We want to follow the spec, and in strict mode (which Sucrase assumes), `this` always has the value `undefined` for a non-method-style function invocation.

A quick script I just used in node to confirm this:
```js
"use strict";
function f() { console.log(this); }
f();
```
which prints "undefined".

aleclarson:
> your code could just "leave a note" in class state and then replace the open-paren in the later call to `CJSImportTransformer.process`.

Are there two calls to each transformer's `process` method every `sucrase.transform` call? Or how should I understand this?

aleclarson:
> Option 2 seems like maybe the best one here. I'll also keep thinking if there are other options.
>
> Oh, and this might be an argument for going back to the _id strategy, since that one is more self-contained syntactically.

I think option 2 is the best... of those options. This is a good argument for conditionally replacing based on existence of closing parens, as I mentioned earlier.

alangpierce:
> Are there two calls to each transformer's process method every sucrase.transform call? Or how should I understand this?

Generally speaking, `Transformer.process` is how a transformer processes a *single* token, though it can decide to process multiple at once if it wants.

The implementation is at `RootTransformer.processToken`, which gets called over and over. Each time, it takes a look at the next token, asks every transformer to process the token if possible, and exits early if a transformer is able to process it. If no transformer is able to process the token, then it just copies the token, which should be the common case. It's a bit delicate because the transformer order defines precedence, so order matters.

What I'm saying is that `CJSImportTransformer` could keep state like `shouldChangeNextOpenParenToCall` or something, then later, `RootTransformer.processToken` will (hopefully) call `CJSImportTransformer.process` for the open-paren token, and `CJSImportTransformer` can do the `(` -> `.call(void 0, ` replacement then. That would allow `processBalancedCode` to do proper bookkeeping of `parenDepth`, since the first `CJSImportTransformer.process` call (processing an identifier) doesn't consume any open or close parens.

alangpierce:
> This is a good argument for conditionally replacing based on existence of closing parens, as I mentioned earlier.

I'm not totally sure, but my impression is that with the `shouldChangeNextOpenParenToCall`, it would still be cleaner and correct to just use the `(` -> `.call(void 0, ` strategy all the time, rather than having two strategies that we pick between. But happy to go either way.

aleclarson:
Take 2 is ready. Apart from `src/transformers/CJSImportTransformer.ts`, all I had to do was add two new arguments to `processBalancedCode` so I can inject the initial state and do recursive processing to clean up the parent state.

aleclarson:
I also have it default to `(0, f)()` if the previous token is an open brace, semicolon, or equal sign. There are probably other valid tokens, so maybe that check can be broken off into its own function.

src/transformers/CJSImportTransformer.ts
```diff
@@ -177,23 +177,48 @@ export default class CJSImportTransformer extends Transformer {
...
+              prevToken.type !== tt.semi &&
+              prevToken.type !== tt.eq
+            ) {
+              // NOTE: This line inserts a new parenthesis, which must be accounted
+              // for by recursively calling the `processBalancedCode` method of the
+              // root transformer.
+              const hasArgs = this.tokens.tokenAtRelativeIndex(2).type !== tt.parenR;
+              this.tokens.replaceToken(`${replacement}.call(` + (hasArgs ? "void 0, " : ""));
+              this.tokens.removeToken(); // Remove the old paren.
+              this.rootTransformer.processBalancedCode(0, 1);
```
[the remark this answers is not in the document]
aleclarson: Ah, very nice.

src/transformers/CJSImportTransformer.ts
```diff
@@ -177,23 +177,48 @@ export default class CJSImportTransformer extends Transformer {
...
+        // by using `(0, f)` when the identifier is in a paren expression. Else
+        // use `Function.prototype.call` when the identifier is a guaranteed
+        // function call. When using `call`, pass undefined as the context.
+        switch (this.tokens.tokenAtRelativeIndex(1).type) {
+          case tt.parenL: {
+            // We can use `(0, f)` when the previous token is an open brace,
+            // semicolon, or equal operator. This saves space for function calls
+            // with arguments, because then `void 0` is omitted.
+            const prevToken = this.tokens.tokenAtRelativeIndex(-1);
+            if (
```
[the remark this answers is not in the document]
aleclarson: Sure thing 👍

src/transformers/CJSImportTransformer.ts
```diff
@@ -177,23 +177,48 @@ export default class CJSImportTransformer extends Transformer {
...
+              // NOTE: This line inserts a new parenthesis, which must be accounted
+              // for by recursively calling the `processBalancedCode` method of the
+              // root transformer.
+              const hasArgs = this.tokens.tokenAtRelativeIndex(2).type !== tt.parenR;
+              this.tokens.replaceToken(`${replacement}.call(` + (hasArgs ? "void 0, " : ""));
+              this.tokens.removeToken(); // Remove the old paren.
+              this.rootTransformer.processBalancedCode(0, 1);
+              return true;
+            }
+          }
```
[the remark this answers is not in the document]
aleclarson: Agreed!

src/transformers/CJSImportTransformer.ts
```diff
@@ -177,23 +177,48 @@ export default class CJSImportTransformer extends Transformer {
...
-      this.tokens.tokens[possibleOpenParenIndex].type === tt.parenR
-    ) {
-      possibleOpenParenIndex++;
-    }
-    if (this.tokens.tokens[possibleOpenParenIndex].type === tt.parenL) {
-      this.tokens.replaceToken(`(0, ${replacement})`);
-    } else {
-      this.tokens.replaceToken(replacement);
+    let openParenIndex = this.tokens.currentIndex();
+    while (++openParenIndex < this.tokens.tokens.length) {
```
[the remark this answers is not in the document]
aleclarson: I'll put it back to the way it was.

aleclarson:
Take 3. Cleaner than ever. LGTM?

src/transformers/CJSImportTransformer.ts
```diff
@@ -177,20 +177,31 @@ export default class CJSImportTransformer extends Transformer {
...
     ) {
       possibleOpenParenIndex++;
     }
+    // Avoid treating imported functions as methods of their `exports` object
+    // by using `(0, f)` when the identifier is in a paren expression. Else
+    // use `Function.prototype.call` when the identifier is a guaranteed
+    // function call. When using `call`, pass undefined as the context.
     if (this.tokens.tokens[possibleOpenParenIndex].type === tt.parenL) {
-      this.tokens.replaceToken(`(0, ${replacement})`);
+      if (this.tokens.tokenAtRelativeIndex(1).type == tt.parenL) {
```
alangpierce: `===` please. (This breaks lint, and you should be able to do `yarn test` locally to catch it.)

Also, maybe you could do `this.tokens.currentIndex() + 1 === possibleOpenParenIndex`?

I *think* you should be able to handle the `new` issues by just making sure the prior token is not `new` in this conditional.

test/imports-test.ts
```diff
@@ -388,11 +388,53 @@ var _moduleName = require('moduleName');
       `"use strict";
       var _mymodule = require('my-module');
       
-      (0, _mymodule.bar)();
+      _mymodule.bar.call(void 0, );
     `,
     );
   });
 
+  it("never breaks weird call expressions", () => {
```
alangpierce: Can you add a new test or an extend one of these tests to include the `new` case?

src/transformers/CJSImportTransformer.ts
aleclarson: Sorry, I live on the edge. Will fix it. (regarding `yarn test`: I've been using `jest -i` with a `jest.config.js` I setup, because I can't stand mocha)

> Also, maybe you could do this.tokens.currentIndex() + 1 === possibleOpenParenIndex?

Nice! At this point, we could just merge it and you fix the nit picks. :P

test/imports-test.ts
aleclarson: 👍

aleclarson:
Test case for `new` expression added.

Now using `(0, f)` for `new` expressions.

aleclarson:
Consider it done

benchmark/sample/expression.ts
```diff
@@ -633,7 +633,7 @@ function parseParenItem(): void {
 // argument to parseSubscripts to prevent it from consuming the
 // argument list.
 function parseNew(): void {
-  parseIdentifier();
+  expect(tt._new);
```
alangpierce: Haha, this one is just a sample TS file for one of the benchmarks, never actually executed, but sure.

test/imports-test.ts
```diff
@@ -388,11 +388,55 @@ var _moduleName = require('moduleName');
...
+    assertResult(
+      `
+      import a from 'a';
+      import b from 'b';
+      (a)();
+      ((a))();
+      a(b)();
+      (a + b)();
+      new (a)();
```
alangpierce: Technically this doesn't exercise your case, since it's surrounded by parens here, but no big deal.

alangpierce:
@aleclarson published in 3.7.1! Thanks for working through all of the details!

aleclarson:
Cheers!

PR #375, opened 2018-12-28
Switch class field implementation from symbols to prototype access

alangpierce:
The previous implementation broke react-hot-loader in two ways:
* When computing the changes to apply to a component instance, it was running
  the old constructor with the new `this.__proto__`. This was crashing because
  the symbols from the old and new files were different, so `this[__init]`
  didn't resolve to a real function.
* When changing from symbols to unique strings, it still broke since RHL was
  expecting the old class fields to be set.

The original motivation for symbols rather than autogenerated names was so that
there couldn't be accidental subclass overriding. An alternate solution is to
get the method directly off of the class and call it rather than following the
prototype chain. This also seems to do what I want with regard to
react-hot-loader behavior and results in simpler emitted code, so seems to be an
improvement all around.

I tested this end-to-end to confirm that bound methods are properly updated with
a later commit that implements the RHL transform.


test/sucrase-test.ts
```diff
@@ -298,10 +298,26 @@ describe("sucrase", () => {
...
+    assertResult(
+      `
+      const A = class {
+        x = 3;
+      }
+    `,
+      `"use strict"; var _class;
+      const A = (_class = class {constructor() { _class.prototype.__init.call(this); }
+        __init() {this.x = 3}
+      }, , _class)
```
alangpierce: Oops, this is buggy...

PR #389, opened 2019-01-03
Properly handle optional arrow params without type parameters

### LINKED ISSUE #387: Optional argument in arrow functions doesn't compile
This doesn't compile with Sucrase 3.9.1 but compiles with TypeScript:

```ts
const test = (a?) => a;
```

See to compare the results:

https://sucrase.io/#compareWithBabel=false&compareWithTypeScript=true&code=%2F%2F%20works%20with%20tsc%2C%20doesn't%20with%20sucrase%0Aconst%20test%20%3D%20(a%3F)%20%3D%3E%20a%3B%0A%0A%2F%2F%20works%20in%20both%20tsc%20and%20sucrase%0Afunction%20test2(a%3F)%20%7B%0A%20%20return%20a%3B%0A%7D

alangpierce:
Fixes #387

I reworked `typedParseConditional` significantly from the Babel implementation
and didn't realize that I needed to preserve bare `?` tokens to denote the
parameter optional. Fortunately, the optional case will always be `?,` or `?)`
(an initializer isn't allowed when the param is optional), and neither of those
cases is valid as the start of a ternary, so we can distinguish optional param
from ternary with only one token of lookahead.

### DIFF
--- a/src/parser/plugins/types.ts
+++ b/src/parser/plugins/types.ts
@@ -12,9 +12,13 @@ import {tsParseTypeAnnotation} from "./typescript";
 // An apparent conditional expression could actually be an optional parameter in an arrow function.
 export function typedParseConditional(noIn: boolean): void {
   // If we see ?:, this can't possibly be a valid conditional. typedParseParenItem will be called
-  // later to finish off the arrow parameter.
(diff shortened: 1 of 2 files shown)


kiliancs:
That was fast! Thank you!

PR #402, opened 2019-01-11
fix: CJSImportTransformer#processAssignment

### LINKED ISSUE #401: TS: Exported generic class with default type parameter
The following snippet..
```ts
export class Foo<T = any> {};
let foo: Foo = new Foo();
let foo: Foo<any> = new Foo();
```
..compiles to..
```ts
"use strict";Object.defineProperty(exports, "__esModule", {value: true}); class Foo {} exports.Foo = Foo;;
let foo = exports.Foo = new Foo();
let foo = new Foo();
```

Notice how `exports.Foo` is being re-assigned. 😳

aleclarson:
Closes #401

### DIFF
--- a/src/transformers/CJSImportTransformer.ts
+++ b/src/transformers/CJSImportTransformer.ts
@@ -274,6 +274,8 @@ export default class CJSImportTransformer extends Transformer {
   private processAssignment(): boolean {
     const index = this.tokens.currentIndex();
     const identifierToken = this.tokens.tokens[index - 1];
+    // If the LHS is a type identifier, this must be a declaration like `let a: b = c;`,
+    // with `b` as the identifier, so nothing needs to be done in that case.
     if (identifierToken.isType || identifierToken.type !== tt.name) {
       return false;
     }
--- a/test/typescript-test.ts
+++ b/test/typescript-test.ts
@@ -1357,7 +1357,7 @@ describe("typescript transform", () => {
     );
   });
 
-  it("fixes issue #401", () => {
+  it("properly handles a declaration that looks like an assignment to an export (#401)", () => {
     assertTypeScriptResult(
       `
       export class Foo {}


src/transformers/CJSImportTransformer.ts
```diff
@@ -274,7 +274,7 @@ export default class CJSImportTransformer extends Transformer {
   private processAssignment(): boolean {
     const index = this.tokens.currentIndex();
     const identifierToken = this.tokens.tokens[index - 1];
-    if (identifierToken.type !== tt.name) {
+    if (identifierToken.isType || identifierToken.type !== tt.name) {
```
alangpierce:
```suggestion
    // If the LHS is a type identifier, this must be a declaration like `let a: b = c;`,
    // with `b` as the identifier, so nothing needs to be done in that case.
    if (identifierToken.isType || identifierToken.type !== tt.name) {
```

test/typescript-test.ts
```diff
@@ -1356,4 +1356,17 @@ describe("typescript transform", () => {
     `,
     );
   });
+
+  it("fixes issue #401", () => {
```
alangpierce:
```suggestion
  it("properly handles a declaration that looks like an assignment to an export (#401)", () => {
```

alangpierce:
@aleclarson drat, almost worked (I think this came up in a previous PR as well). Would you mind accepting the suggestions? (Also happy to discuss tweaks.)

screen shot 2019-01-12 at 10 15 26 am

aleclarson:
Good to go 👍

aleclarson:
Have you tried [semantic-release](https://github.com/semantic-release/semantic-release) before? I use it on all of my repos nowadays. 😄

alangpierce:
Yep, I have used it before! I was meaning to get something like that set up for Sucrase, but after the eslint-scope fiasco ( https://eslint.org/blog/2018/07/postmortem-for-malicious-package-publishes ), I decided that I need two-factor auth for npm publishes (`auth-and-writes`), which doesn't work with semantic-release ( https://github.com/semantic-release/npm#npm-registry-authentication ) or any system that publishes from CI.

I also like having a changelog that's a little more human-curated rather than just all of the commit message titles.

I am certainly on-board with frequent publishes. I'll do a publish pretty soon, definitely this weekend, I just want to see if I can get a few more fixes in.

PR #414, opened 2019-02-16
Update README, including new swc perf comparison

### LINKED ISSUE #413: I made a broccoli plugin
Hey Cool library, this isn't an issue but I just wanted to share I created a broccoli plugin for sucrase: 
https://github.com/stefanpenner/broccoli-sucrase

alangpierce:
I installed swc and hacked the benchmark to run it on the codebase used for the
comparison (the Benchling frontend codebase). There were about 100 places in the
code that crashed the swc parser, so I commented those out or otherwise modified
the code to get it working. I ran swc 10 times and took the fastest time, which
was 4.7x slower than Sucrase. One caveat is that swc was targeting ES3, since it
doesn't have an option to target latest JS.

Also mention the Broccoli integration, fixes #413.

Also fix a few other copy issues in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,22 +29,23 @@ the [Benchling](https://benchling.com/) frontend code,
 with all tests passing, about 1 million lines of code total.
 
 **Sucrase is about 20x faster than Babel.** Here's one measurement of how Sucrase
-compares with tsc and Babel on a large TypeScript codebase with 4045 files and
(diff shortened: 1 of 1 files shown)


Hammster:
The benchmark in its current state is very misleading given that you run SWC to compile to ES3.

You can actually set the target via https://swc-project.github.io/docs/configuring-swc#jsctarget since v1.0.27 too.

PR #425, opened 2019-02-28
Allow component names to start with _

### LINKED ISSUE #424: <_Foo/> treated as tag rather than component
`<_Foo/>` is treated as a tag type rather than component type, so a string is emitted to createElement.

Also: sucrase is awesome!

yang:
Fixes #424

### DIFF
--- a/src/transformers/JSXTransformer.ts
+++ b/src/transformers/JSXTransformer.ts
@@ -3,6 +3,7 @@ import {Options} from "../index";
 import NameManager from "../NameManager";
 import XHTMLEntities from "../parser/plugins/jsx/xhtml";
 import {TokenType as tt} from "../parser/tokenizer/types";
+import {charCodes} from "../parser/util/charcodes";
 import TokenProcessor from "../TokenProcessor";
 import getJSXPragmaInfo, {JSXPragmaInfo} from "../util/getJSXPragmaInfo";
 import RootTransformer from "./RootTransformer";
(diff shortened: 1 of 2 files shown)


src/transformers/JSXTransformer.ts
```diff
@@ -252,7 +252,7 @@ export default class JSXTransformer extends Transformer {
 }
 
 export function startsWithLowerCase(s: string): boolean {
-  return s[0] === s[0].toLowerCase();
+  return s[0] === s[0].toLowerCase() && s[0] !== "_";
```
alangpierce: `$` is another identifier character that can come up here: [playground link](https://sucrase.io/#code=%0A<%24><%2F%24>)

I think if you invert it, that should catch it rather than having to list special cases. So this should work:
```typescript
return s[0] !== s[0].toUpperCase()
````

Comment (yang) [reply src/transformers/JSXTransformer.ts]: 👍 

BTW was there somewhere you looked to know about the `$` thing?

Comment (alangpierce) [reply src/transformers/JSXTransformer.ts]: Nowhere I looked, just remembering the first time I saw jQuery code and thinking "how is that possibly valid?", then learning that `$` is legal in JS identifiers.

But I do have some familiarity with the spec now. It's defined here:
https://tc39.github.io/ecma262/#prod-IdentifierStart

Another thing that the spec reminded me of is that it also needs to work for non-ASCII. Turns out `<é />` is also buggy ([playground](https://sucrase.io/#compareWithTypeScript=true&code=%0A<é%20%2F>)). It is *not* lowercase according to JSX rules, even though `'é'.toUpperCase() === 'É'`. So I think it's really just lowercase "a" through "z" that make the `createElement` argument a string.

So actually probably best to do something like this:
```typescript
return s.charCodeAt(0) >= charCodes.lowercaseA && s.charCodeAt(0) <= charCodes.lowercaseZ;
```

(`charcodes.ts` is an existing file that lives in `src/parser/util`)

alangpierce:
@yang tests should be fixed now! Can you rebase? Also would be good to switch to the `>= charCodes.lowercaseA && <= charCodes.lowercaseZ` approach I mentioned, which also would likely be a little faster since it doesn't need to handle unicode.

alangpierce:
@yang Thanks, I just published as 3.9.6.

PR #448, opened 2019-05-04
Support .tsx and .jsx files from CLI

ricardobeat:
Despite the transforms being supported, the CLI will currently only compile `.ts` files. This makes sure the correct file extensions are used when running something like `sucrase ./src -d ./out --transforms imports,typescript,jsx`.


src/cli.ts
```diff
@@ -87,10 +87,8 @@ async function buildDirectory(
     const outChildPath = join(outDirPath, child);
     if ((await stat(srcChildPath)).isDirectory()) {
       await buildDirectory(srcChildPath, outChildPath, options);
-    } else if (srcChildPath.endsWith(extension)) {
-      const outPath = `${outChildPath.substr(0, outChildPath.length - extension.length)}.${
-        options.outExtension
-      }`;
+    } else if (extension.test(srcChildPath)) {
+      const outPath = _path.basename(outChildPath, _path.extname(outChildPath)) + '.' + options.outExtension;
```
alangpierce: I'm getting `ReferenceError: _path is not defined` when I run this. Maybe you meant to import it?

ricardobeat: Fixed imports.

ricardobeat:
Fixed - the other PR is almost exactly the same, only uses a regexp instead of `path` to handle the extension. Apologies for the mistake - for some reason `yarn build` only works in the master branch, though tests are now passing.

alangpierce:
Thanks!

Looks like it was still having issues because it wasn't the full path, just the filename portion. I switched to the regex approach in the other PR, unclear which one is "best", but certainly the regex is a bit more concise.

PR #453, opened 2019-05-29
Remove matches+Number, use spread syntax

sh7dm:
Refactor, reduces code size, make development easier

### DIFF
--- a/src/CJSImportProcessor.ts
+++ b/src/CJSImportProcessor.ts
@@ -54,14 +54,14 @@ export default class CJSImportProcessor {
   preprocessTokens(): void {
     for (let i = 0; i < this.tokens.tokens.length; i++) {
       if (
-        this.tokens.matches1AtIndex(i, tt._import) &&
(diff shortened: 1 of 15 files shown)


alangpierce:
Thanks for the PR! Unfortunately, I don't think I'll be able to accept it.

I actually originally had a single `matches` function like this, then split it out into separate functions for performance reasons. It's one of the most heavily called functions, so optimizing function call overhead is important, and JS engines have a harder time optimizing invocations of variadic functions or functions accepting arrays.

Here's what I get when I run `yarn benchmark` 5 times without your change applied:
```
Sucrase: 451.404ms
Sucrase: 449.287ms
Sucrase: 438.189ms
Sucrase: 443.050ms
Sucrase: 447.706ms
```

Here's what I get when I run it 5 times with your change applied:
```
Sucrase: 531.466ms
Sucrase: 531.840ms
Sucrase: 526.705ms
Sucrase: 528.225ms
Sucrase: 532.726ms
```

So even though `matches1`, `matches2`, etc are a bit ugly, their performance benefits are worth it, I think, especially for a performance-focused project like this.

PR #458, opened 2019-06-05
add jsx input ext

### LINKED ISSUE #456: support customizing input extensions. e.g. jsx
sucrase currently ignores .jsx files
I don't want to mix a js and jsx files into .js ext

a-x-:
closes #456 

or, it might be better to add  --exts otp

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -87,10 +87,8 @@ async function buildDirectory(
     const outChildPath = join(outDirPath, child);
     if ((await stat(srcChildPath)).isDirectory()) {
       await buildDirectory(srcChildPath, outChildPath, options);
-    } else if (extension.some((ext) => srcChildPath.endsWith(ext))) {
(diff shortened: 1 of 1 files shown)


src/cli.ts
```diff
@@ -75,7 +75,9 @@ async function buildDirectory(
   outDirPath: string,
   options: CLIOptions,
 ): Promise<void> {
-  const extension = options.sucraseOptions.transforms.includes("typescript") ? ".ts" : ".js";
+  const extensions = options.sucraseOptions.transforms.includes("typescript")
+    ? [".ts"]
```
alangpierce: Could you add `.tsx` here as well?

a-x-: yeah, I thinked about and I was not sure

alangpierce:
#448 just landed, so closing this out. I'll give both of you credit in the release notes.

PR #463, opened 2019-06-25
Do not remove generator markers

### LINKED ISSUE #462: Generator methods are not transpiled correctly
Even with no transformers enabled, generator methods are converted to normal methods, likely leading to an error on the yield keyword.
```js
class Foo {
  *generator() {
    yield 32;
  }
}
```
is transpiled to
```js
class Foo {
  generator() {
    yield 32;
  }
}
```

itsbth:
Fixes #462 

Previously, the star ended up in rangesToRemove.
Considered simply adding it to isAccessModifier, but I don't really think it is one.


itsbth:
Moved it to isAccessModifier. Didn't rename the function, but can do that if you want that in this PR.

test/sucrase-test.ts
```diff
@@ -249,6 +249,24 @@ describe("sucrase", () => {
...
+      `"use strict";
+      class A {
+        *foo() {
+        }
+      }
+    `,
+      {transforms: ["jsx", "imports", "typescript"]},
+    );
+  });
+
```
arv: Can you also add a test for?

```js
class C {
  async *m() {
    yield await 1;
  }
}
```

arv:
Thanks

Issue #507, opened 2020-02-12
Can we run sucrase in the browser like babel-standalone?


alexilyaev:
For the same reasons [babel-standalone](https://babeljs.io/docs/en/babel-standalone) exists.
Would be nice to be able to run sucrase.

Right now babel-standalone doesn't support React Fragments (it's stuck on babel v7 beta 3).

alangpierce:
Hey @alexilyaev , Sucrase itself runs fine in the browser (it doesn't have any node dependencies), though at the moment there's no pre-packaged browser build attaching it to the global scope, and there's no support for `text/sucrase`-style scripts.

In the past, I've been hesitant about a browser bundle the attaches Sucrase as a global since it's kind of the "old way" of using libraries, though the `text/sucrase` type of use case does seem compelling. In your case you'd just use the jsx transform, is that right? Seems like the typescript, flow, and imports transforms would only make sense in environments with additional tooling where you might as well have webpack (or similar) set up, though I could see a use case for TS/Flow checking and still wanting the script directly loaded to the browser rather than through webpack.

alexilyaev:
Well, my use case is teaching React with no bundler, currently it looks like this:

```html
<div id="root"></div>

<script src="https://unpkg.com/react@16.12.0/umd/react.development.js"></script>
<script src="https://unpkg.com/react-dom@16.12.0/umd/react-dom.development.js"></script>
<script src="https://unpkg.com/babel-standalone@6.26.0/babel.js"></script>

<script type="text/babel" src="app.js"></script>
```
```js
// app.js
const App = () => {
  return <h1>Hello World!</h1>;
};

ReactDOM.render(<App />, document.querySelector('#root'));
```

So React is global.
`import` and `require` don't work in this context.
TypeScript would have been a nice addition as it useful for teaching.

Right now I'm only missing the React Fragment `<>...</>` syntax support, which was added in `babel-standalone v7.0.0-beta.31`...

### EDIT:

I was using an old package of babel-standalone, the new one has the latest versions:
https://unpkg.com/browse/@babel/standalone@7.8.4/

So this one does support the React Fragment syntax (and TS as well).

```html
<script src="https://unpkg.com/@babel/standalone@7.8.4/babel.min.js"></script>
```

But, that file weighs 1.6 MB.
So, if `sucrase` could do it leaner, I'd be happy about that.

tomByrer:
duplicate of #506 

[...]

PR #526, opened 2020-04-20
add browser build with rollup

### LINKED ISSUE #447: Publishing on CDN
Sometimes I don't setup the project with `package.json` but just use `<script>`'s. And when I want to write a small app with React **I have to include Babel**, but it is very slow. I'd like to **include Sucrase instead**.

The most popular CDNs are `unpkg.com` and `cdnjs.com`.

I hope it won't be difficult to do.

### LINKED ISSUE #506: please release bundled code
Sorry to ask for the extra work.  Should help fix #447 since jsDeliver etc will automatically host your code.  Right now, it seems to be a CLI, so I'm unsure how to compile.  I guess I could point ParcelJS at thei repo, but I'm unsure what to do if something doesn't work or if it doesn't work as intended...

TIA :)

nitsky:
Hi Alan, thanks for your work on sucrase. It has sped up incremental compilation of the app I work on from 4 sec to 400 ms. That is the difference between getting annoyed each time I hit save and having the build finish faster than I can switch from my editor to my browser. Thanks!

I can currently import rollup directly in a modern browser like so:

```js
import * as rollup from 'https://unpkg.com/rollup@2.3.3/dist/es/rollup.browser.js'
```

I would like to be able to do the same with sucrase:

```js
import * as sucrase from 'https://unpkg.com/sucrase@3.13.0/dist/sucrase.browser.js'
```

This PR adds a basic rollup config and emits sucrase.browser.js to the dist directory when a build is run.

This browser build does not attach sucrase to the global scope and does not process text/sucrase scripts, as suggested in #507, but it closes #447 and closes #506.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,8 @@
     "@babel/preset-flow": "^7.0.0",
     "@babel/preset-react": "^7.7.0",
     "@babel/preset-typescript": "^7.7.2",
+    "@rollup/plugin-commonjs": "^11.1.0",
+    "@rollup/plugin-node-resolve": "^7.1.3",
     "@types/glob": "7.1.1",
     "@types/mocha": "^5.2.7",
     "@types/mz": "^0.0.32",
@@ -71,6 +73,7 @@
     "mocha": "^7.1.1",
     "nyc": "^14.1.1",
     "prettier": "^1.19.1",
+    "rollup": "^2.6.1",
     "sucrase": "^3.13.0",
     "test262-harness": "^6.5.0",
     "ts-interface-builder": "^0.2.1",
@@ -85,6 +88,7 @@
     "lines-and-columns": "^1.1.6",
     "mz": "^2.7.0",
     "pirates": "^4.0.1",
+    "rollup-plugin-terser": "^5.3.0",
     "ts-interface-checker": "^0.1.9"
   },
   "engines": {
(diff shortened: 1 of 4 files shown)


alangpierce:
Hi @nitsky, thanks for the contribution, and sorry for the delay in getting to this.

I just tried it out locally and it definitely works and is clean. The main thing I want to make sure is that it doesn't preclude future use cases. I'm not really that worried since other use cases can always be hosted under different filenames, but seems worth a little discussion before publishing.

Could you explain a little further how you intend to use this? Is it intended for the "demo website" use case (like https://sucrase.io/ ), or are you planning to use Sucrase to compile your own code in-browser? The `text/sucrase` use case has felt the most compelling to me because it lowers the barrier to entry, though I think it only really makes sense in an educational context since it doesn't let you split your code up across multiple files. When I looked before, I didn't find a way to hook into the browser's `import` behavior to run Sucrase on-the-fly, but maybe with rollup you could do cross-file in-browser compilation?

Sorry for the build break. I fixed it recently, so if you merge/rebase, the build should pass I think.

nitsky:
@alangpierce I am compiling a web application with rollup and sucrase from a [deno](https://github.com/denoland/deno) script as opposed to node. After submitting this PR I discovered that sucrase can be conveniently imported from the [pika](https://cdn.pika.dev) cdn like so:

```js
import * as sucrase from 'https://cdn.pika.dev/sucrase@3.14.0'
```

This has satisfied my needs, so you can close this PR if you wish.

ClassicOldSong:
I don't think this PR should be closed since there's still use case that requires UMD or IIFE in Service Workers.

If there's no intention for @alangpierce to have this implemented, I'll open a new package that only compiles Sucrase into UMD and IIFE formats for this convenience,  or we should have a new PR which does the exact same thing.

I'd like to hear the author's opinion.

PR #533, opened 2020-05-16
Add support for TS type-only re-exports

Rugvip:
Hi! I ran into some trouble using type re-exports in TypeScript.

```ts
export type { T } from './T';
```

would be transpiled to

```ts
       from './T';
```

The syntax form is mentioned in the PR adding the functionality to TS: https://github.com/microsoft/TypeScript/pull/35200

Wondering if this looks like an alright fix?


alangpierce:
@Rugvip I just released this as 3.14.1!

Issue #539, opened 2020-05-27
Hoist Jest mocks


Rugvip:
I tried using Sucrase to speed up Jest tests, and ran into issues with module mocks not being hoisted up above imports. Without automatic hoisting, the `jest.mock`/`jest.unmock` calls need to be manually placed above the imports, which isn't optimal.

This is for example how `ts-jest` does hoisting: https://github.com/kulshekhar/ts-jest/blob/a6bcec48ef28791235b3eba0b3f2bd1a944ba5b8/src/transformers/hoist-jest.ts

Would be awesome to have this implemented! I could have a look at it as well, given some pointers.

Raynos:
That's called `JestScript`

If your changing the order of execution of javascript statements in a file that ends with `.js` then it's not JavaScript.

There are transforms for `jsx` & `flow` ; I guess you can write a transformer and add it to https://github.com/alangpierce/sucrase/tree/master/src/transformers

Rugvip:
Fixed by #540

PR #540, opened 2020-05-31
Add jest transform for hoisting module mocks

Rugvip:
Made an attempt at #539, let me know if this makes sense :grin:

It doesn't do all the same checks as babel-plugin-jest-hoist, which verifies that only whitelisted out-of-scope variables are accessed inside module mock factories.

Also had to add `getHoistedCode` to `Transformer`, to be able to move code up below all the helpers. There might be some nicer way to do this though.

It also doesn't support non-imports transform + jest, but since jest currently doesn't support ESM, that can be added later. Figured the most straight-forward way to do that would be to add global names to the `SucraseContext`, regardless of which transform is selected. Basically adding an equivalent of `importProcessor.getGlobalNames()` and `getTSImportedNames(tokenProcessor)` for when neither `imports` or `typescript` processors are active.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,9 @@ transforms are available:
   transform in the [react-hot-loader](https://github.com/gaearon/react-hot-loader)
   project. This enables advanced hot reloading use cases such as editing of
   bound methods.
+* **jest**: Hoist desired [jest](https://jestjs.io/) method calls above imports in
+  the same way as [babel-plugin-jest-hoist](https://github.com/facebook/jest/tree/master/packages/babel-plugin-jest-hoist).
+  Does not validate the arguments passed to `jest.mock`, but the same rules still apply.
 
 These proposed JS features are built-in and always transformed:
 * [Optional chaining](https://github.com/tc39/proposal-optional-chaining): `a?.b`
--- a/integrations/jest-plugin/README.md
+++ b/integrations/jest-plugin/README.md
@@ -24,5 +24,5 @@ Then add it as a transform to your Jest config in package.json:
 ```
 
 Currently, the transforms are not configurable; it uses always runs the import
-transform and uses the file extension to decide whether to run the JSX, Flow,
-and/or TypeScript transforms.
+and jest transforms and uses the file extension to decide whether to run the
+JSX, Flow, and/or TypeScript transforms.
--- a/integrations/jest-plugin/src/index.ts
+++ b/integrations/jest-plugin/src/index.ts
@@ -2,11 +2,11 @@ import {Transform, transform} from "sucrase";
 
 function getTransforms(filename: string): Array<Transform> | null {
   if (filename.endsWith(".js") || filename.endsWith(".jsx")) {
-    return ["flow", "jsx", "imports"];
+    return ["flow", "jsx", "imports", "jest"];
   } else if (filename.endsWith(".ts")) {
-    return ["typescript", "imports"];
+    return ["typescript", "imports", "jest"];
   } else if (filename.endsWith(".tsx")) {
-    return ["typescript", "jsx", "imports"];
+    return ["typescript", "jsx", "imports", "jest"];
   }
   return null;
 }
--- a/src/Options-gen-types.ts
+++ b/src/Options-gen-types.ts
@@ -10,6 +10,7 @@ export const Transform = t.union(
   t.lit("flow"),
   t.lit("imports"),
   t.lit("react-hot-loader"),
+  t.lit("jest"),
 );
 
 export const SourceMapOptions = t.iface([], {
--- a/src/Options.ts
+++ b/src/Options.ts
@@ -3,7 +3,7 @@ import OptionsGenTypes from "./Options-gen-types";
 
 const {Options: OptionsChecker} = createCheckers(OptionsGenTypes);
 
-export type Transform = "jsx" | "typescript" | "flow" | "imports" | "react-hot-loader";
+export type Transform = "jsx" | "typescript" | "flow" | "imports" | "react-hot-loader" | "jest";
 
 export interface SourceMapOptions {
   /**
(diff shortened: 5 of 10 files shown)


freben:
@alangpierce is this something that looks reasonable to you?

freben:
Ping :) This is still something that we would value.

SebDuf:
I've just tested this branch and it fixes all the problems I had with the current jest plugin and all my tests pass, it would be great for this to be merged!

RIP21:
It's been a while folks! Please merge this one :)

PatrykMilewski:
@alangpierce Any chance to release it?

alangpierce (approved):
Thanks so much for the PR, and sorry for the delay in getting to this! I had been putting it off compared with other Sucrase priorities, and have had a lot of other things going on in my life and have had only occasional bursts of focus time for Sucrase. I've had a few lingering hesitations about this PR (more details below), but after thinking it through more, **I think it makes the most sense to merge and release this, since it seems like a safe change and clearly would provide value**.

I really appreciate the documentation, tests, thoroughness, and general code quality. I found some small issues that I can fix in a quick follow-up PR.

To explain some of my higher-level hesitations (which ideally I would have surfaced sooner):

### The PR breaks line number preservation when the `jest` transform is enabled.

Sucrase so far has had a guarantee that line numbers are the same before and after transformation, which makes source maps trivial because each line just maps to itself. Because this Jest transform reorders multi-line blocks, it ends up shifting line numbers for most/all of the file, which means that debugger breakpoints won't work correctly.

This is generally a challenging problem with any transform related to reordering, and I remember running into this with the class fields transform as well. I originally implemented it by moving the initializer code to the constructor, but that shifted line numbers. One possible workaround is to join the code snippet to one long line, but that raises some worries around automatic semicolon insertion and needs some complexity around multiline string literals and `//` comments. My eventual solution for class fields was to wrap each initializer in a method declaration and then call the relevant method.

An analogous approach here may be to transform this code:
```ts
import 'moduleName';
jest.mock('a', () => {
  console.log('Hi');
});
jest.mock('b');
```

into this:
```ts
__init1(); __init2(); import 'moduleName';
function __init1() { jest.mock('a', () => {
  console.log('Hi');
}); }
function __init2() { jest.mock('b'); }
```

It's a bit ugly in that it's relying on JS hoisting for `function` declarations, but I think it should work. It does add some name generation and lookahead complexity to the transform, though, so I think it's fine to release this PR and optionally fix the line number issue in the future.

### I don't personally have much experience with Jest.

Sucrase and my employer both use Mocha, so I haven't had as much hands-on experience with Jest and wasn't previously aware of the hoisting issue. I feel a little embarrassed since it's such a widely-used tool, and have been meaning to understand it more deeply. A few months ago I tried switching the Sucrase test suite to use Jest, but the test suite time slowed down from about half a second to about 5 seconds (I believe due to the sandboxing that Jest does), so I decided to stick with Mocha for now. What that means in practice here is that **I haven't personally tested this PR in a realistic setting, so I'm relying on the community to ensure quality and report issues**.

### It's unclear if this transform should be necessary in the first place.

To me, it feels a little strange to write `.mock` calls below imports and then expect the mocking to happen before the import runs. Babel reorders import statements to the top of the file, so it makes sense that in a Babel setup you'd need an *additional* transform to reorder the `.mock` calls even higher. Sucrase, however, follows the TypeScript approach of leaving `import` statements where they are, so a `.mock` call before an `import` statement will preserve that order.

That said, I understand that `.mock` after imports may be more idiomatic or may play better with editors and other tools. It just feels a little strange to me that this is solved using a code transform.

### It's unclear if this transform should be considered in the scope of Sucrase.

Sucrase is generally concerned with non-standard syntax extensions (particularly JSX and TypeScript), though there are certainly exceptions to that. It's probably too big of a scope for Sucrase to support *every* tool that needs a custom code transform, but it does seem pragmatic to pick the most important tools to support, which was the rationale for the `react-hot-loader` transform. It's probably fair to count Jest as one of the most important tools to support, but I think generally each tool needs to be decided case-by-case.

----

To answer a question in the PR summary:

> It also doesn't support non-imports transform + jest, but since jest currently doesn't support ESM, that can be added later. Figured the most straight-forward way to do that would be to add global names to the SucraseContext, regardless of which transform is selected. Basically adding an equivalent of importProcessor.getGlobalNames() and getTSImportedNames(tokenProcessor) for when neither imports or typescript processors are active.

I agree that it seems reasonable to skip ESM support for this PR. My main hesitation with adding more information to the SucraseContext by default is the performance cost of computing globals when they aren't needed, so that would be the main thing to be cautious of, but overall the code strategy looks right to me.

RIP21:
@alangpierce BTW. Is Surcrase supports react-refresh? (as you mentioned it support HMR, but I'm unsure react-refresh and HMR are connected)

RIP21:
@alangpierce and another question. I guess this will be released after you or somebody else fixes the source map issue? As I guess broken debugger is kinda meh and I would rather hoist myself for the time being than using this transform but keep the debugger experience perfect as it was before :) 

Thanks! :)

alangpierce:
> I guess this will be released after you or somebody else fixes the source map issue?

My plan was actually to release this with the broken line numbers, since it only affects people explicitly using the `jest` transform. But let me see if I can get my code transformation above working now and include it as part of the release.

alangpierce:
I just put up a line number fix at https://github.com/alangpierce/sucrase/pull/608! I did some basic testing, but if anyone has a chance to test it on a real codebase, would be great to hear any feedback or issues. If not, I can do a release soon and any potential issues can be fixed in a follow-up release.

alangpierce:
Just published as `sucrase` 3.18.0 and `@sucrase/jest-plugin` 2.1.0!

RIP21:
@alangpierce awesome! Will check tomorrow :)

PR #553, opened 2020-09-07
updating doc jest-plugin

oliveira-andre:
the change on package.json doesn't work for me, then i search a new way to configure sucrase on jest, then i found the transform on jest config file and it works, then i updated the doc to the new way of configuring the jest transform.


alangpierce (approved):
Thanks!

PR #556, opened 2020-09-23
Port babel-parser changes from 2020-04-12 to 2020-07-22

Rugvip:
a466f 5 months ago fix: report missing plugins on type exports (#11417)
🚫 Looks like this is already covered by the existing isTS and isFlow cases

fba64 5 months ago fix: disallow expression after binding identifier `of` (#11355)
🚫 Sucrase doesn't use token contexts.

40c51 5 months ago Set exprAllowed to false for star token (#11449)
🚫 Sucrase doesn't use token contexts.

2e4f18 5 months ago Add some parser missing plugins errors (#11478)
🚫 Validation related to babel plugins only.

fa98a 5 months ago docs: update AST spec (#11492)
🚫 Docs update

9c284 5 months ago (tag: v7.9.6) v7.9.6
🚫 Release only

90a91 5 months ago Update Flow to 0.123.0 (#11500)
🚫 Flow version bump with one change to the tokenizer that isn't present in sucrase

31b36 5 months ago Use ?. where it represents the intended semantics (#11512)
🚫 Refactor to make use of ?. in the all of babel, separate effort in sucrase if needed

2f31e 4 months ago fix: allow bigInt in method name and TSLiteralType (#11547)
✅ Skipped refactor of isLiteralPropertyName in utils, but ported bigint type literal support + added test

62e68 4 months ago Fix comments for smartPipeline topic-forbidding contexts (#11597)
🚫 Comment update related to contexts

5dd7f 4 months ago Enable `import.meta` by default in `@babel/parser` (#11406)
🚫 Import meta is already always parsed

66b86 4 months ago added basic support for module attributes and tests updated (#10962)
🚫 Skipped as this is the implementation of a stage 1 proposal - the proposal is now stage 3 though, but is perhaps a separate effort? https://github.com/tc39/proposal-import-assertions

74590 4 months ago Add private-property-in-object support (#11372)
✅ Ported

bda75 4 months ago Handle private access chained on an optional chain (#11248)
✅ Ported

5da24 4 months ago (tag: v7.10.0) v7.10.0
🚫 Release only

242d9 4 months ago Use `repository.directory` field in `package.json` files (#11625)
🚫 Just babel package.json stuff

88f57 4 months ago (tag: v7.10.1) v7.10.1
🚫 Release only

b5c4a 4 months ago refactor: split locationParser into ParserErrors and error message (#11653)
🚫 Skipped, error message related refactoring

b0350 4 months ago (tag: v7.10.2) v7.10.2
🚫 Release only

71d352 4 months ago Properly parse `export default from` when `exportDefaultFrom` is not enabled (#11676)
✅ Ported the change, but was likely working already.

41085 4 months ago Update prettier to v2 (#11579)
🚫 Skipped, formatting change

b27ab 3 months ago fix: add optional: false to MemberExpression (#11709)
🚫 estree not supported.

e15a5 3 months ago Fix innercomments (#11697)
🚫 Sucrase skips comments

eea15 3 months ago Migrate from "master" branch to "main" (#11715)
🚫 Branch rename

2787e 3 months ago (tag: v7.10.3) v7.10.3
🚫 Release only

30835 3 months ago fix: implement early errors for record and tuple (#11652)
🚫 Error handling only

beca7 3 months ago Add better parser error when using jsx (#11722)
🚫 Error handling only

75c23 3 months ago Add @babel/eslint-plugin-development-internal (#11376)
🚫 Linter changes only

7fd40d 3 months ago (tag: v7.10.4) v7.10.4
🚫 Release only

b1b21 3 months ago docs: add AST spec on optional chain [skip ci] (#11729)
🚫 Documentation change only

d6762 3 months ago fix: throw expect jsx plugin error when an idStart or > is seen (#11774)
🚫 Error handling only

02c8f 3 months ago fix: add optional: false to chained optional call expression (#11814)
❓ Sets node.optional, assuming this can be skipped since optional chaining is handled differently in Sucrase?

f7964 2 months ago (tag: v7.10.5) v7.10.5
🚫 Release only

8f191 10 weeks ago chore: fix typo in codebase (#11846)
🚫 These typos aren't present in Sucrase

f4eeff 10 weeks ago fix: correctly set innerEndPos in CoverParenthesizedExpressionAndArrowParameterList (#11847)
🚫 Fix for cone that's not present in Sucrase

3680f 9 weeks ago fix: allow 09.1_1 and 09e1_1 in sloppy mode (#11854)
🚫 Sucrase doesn't have separate handling of octals so this was not needed

2bf38 9 weeks ago fix: disallow \8, \9 in strict mode string (#11852)
🚫 Error handling only

059e91 8 weeks ago Add decimal parsing support (#11640)
✅ Ported

d7347f 8 weeks ago eslint-parser: ES2020 features (#11815)
🚫 estree not supported.

5b4b3 3 months ago TypeScript 4.0: Allow spread in the middle of tuples (#11753)
✅ Added tests, but the change did not need to be ported

8a1d7e4 3 months ago Allow unknown/any in TS catch clause param (#11755)
✅ Ported + added TypeScript tests, only change needed to support catch clause params

9e666 3 months ago Follow-up on initial TS4 catch param support (#11767)
🚫 Skipped, this detail is not included in the TypeScript parser in sucrase

eba4c 2 months ago TypeScript 4.0: Support labeled tuple elements (#11754)
✅ Ported, excluding validation logic

0e985 9 weeks ago feat: enable numericSeparator parsing support (#11863)
🚫 Skipped, already supported

b651a 9 weeks ago Enable logical assignment by default in @babel/parser (#11860) (#11869)
🚫 Skipped, already supported

### DIFF
--- a/src/parser/plugins/typescript.ts
+++ b/src/parser/plugins/typescript.ts
@@ -367,6 +367,7 @@ function tsParseMappedType(): void {
 function tsParseTupleType(): void {
   expect(tt.bracketL);
   while (!eat(tt.bracketR) && !state.error) {
+    // Do not validate presence of either none or only labeled elements
     tsParseTupleElementType();
     eat(tt.comma);
   }
@@ -376,11 +377,17 @@ function tsParseTupleElementType(): void {
   // parses `...TsType[]`
   if (eat(tt.ellipsis)) {
     tsParseType();
-    return;
+  } else {
+    // parses `TsType?`
+    tsParseType();
+    eat(tt.question);
   }
-  // parses `TsType?`
-  tsParseType();
-  eat(tt.question);
+
+  // The type we parsed above was actually a label
+  if (eat(tt.colon)) {
+    // Labeled tuple types must affix the label with `...` or `?`, so no need to handle those here
+    tsParseType();
+  }
 }
 
 function tsParseParenthesizedType(): void {
--- a/src/parser/plugins/typescript.ts
+++ b/src/parser/plugins/typescript.ts
@@ -413,6 +413,7 @@ function tsParseNonArrayType(): void {
     case tt.string:
     case tt.num:
     case tt.bigint:
+    case tt.decimal:
     case tt._true:
     case tt._false:
       parseLiteral();
--- a/src/parser/plugins/typescript.ts
+++ b/src/parser/plugins/typescript.ts
@@ -412,6 +412,7 @@ function tsParseNonArrayType(): void {
       return;
     case tt.string:
     case tt.num:
+    case tt.bigint:
     case tt._true:
     case tt._false:
       parseLiteral();
--- a/src/parser/tokenizer/index.ts
+++ b/src/parser/tokenizer/index.ts
@@ -4,7 +4,7 @@ import {input, isFlowEnabled, state} from "../traverser/base";
 import {unexpected} from "../traverser/util";
 import {charCodes} from "../util/charcodes";
 import {IS_IDENTIFIER_CHAR, IS_IDENTIFIER_START} from "../util/identifier";
-import {IS_WHITESPACE} from "../util/whitespace";
+import {IS_WHITESPACE, skipWhiteSpace} from "../util/whitespace";
 import {ContextualKeyword} from "./keywords";
 import readWord from "./readWord";
 import {TokenType, TokenType as tt} from "./types";
@@ -216,6 +216,20 @@ export function lookaheadTypeAndKeyword(): TypeAndKeyword {
   const contextualKeyword = state.contextualKeyword;
   state.restoreFromSnapshot(snapshot);
   return new TypeAndKeyword(type, contextualKeyword);
+}
+
+export function nextTokenStart(): number {
+  return nextTokenStartSince(state.pos);
+}
+
+export function nextTokenStartSince(pos: number): number {
+  skipWhiteSpace.lastIndex = pos;
+  const skip = skipWhiteSpace.exec(input);
+  return pos + skip![0].length;
+}
+
+export function lookaheadCharCode(): number {
+  return input.charCodeAt(nextTokenStart());
 }
 
 // Read a single token, updating the parser object's token-related
--- a/src/parser/tokenizer/index.ts
+++ b/src/parser/tokenizer/index.ts
@@ -763,13 +763,17 @@ function readInt(): void {
 
 function readRadixNumber(): void {
   let isBigInt = false;
+  const start = state.pos;
 
   state.pos += 2; // 0x
   readInt();
 
-  if (input.charCodeAt(state.pos) === charCodes.lowercaseN) {
+  const nextChar = input.charCodeAt(state.pos);
+  if (nextChar === charCodes.lowercaseN) {
     ++state.pos;
     isBigInt = true;
+  } else if (nextChar === charCodes.lowercaseM) {
+    unexpected("Invalid decimal", start);
   }
 
   if (isBigInt) {
@@ -783,6 +787,7 @@ function readRadixNumber(): void {
 // Read an integer, octal integer, or floating-point number.
 function readNumber(startsWithDot: boolean): void {
   let isBigInt = false;
+  let isDecimal = false;
 
   if (!startsWithDot) {
     readInt();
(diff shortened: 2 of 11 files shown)


alangpierce (changes requested):
Great, thanks! The one that you ported looks good. For the three where you had questions, I think the private fields ones should be ported as well so that the parser stays up-to-date with future JS.

> 66b86 4 months ago added basic support for module attributes and tests updated (#10962)
> ❓ Skipped as this is the implementation of a stage 1 proposal - the proposal is now stage 3 though, but is perhaps a separate effort? https://github.com/tc39/proposal-import-assertions

Makes sense to skip given that the syntax has changed since that implementation. Whenever babel-parser implements the new `assert` syntax, the port of that commit would be a good opportunity to implement the syntax in Sucrase.

> 74590 4 months ago Add private-property-in-object support (#11372)
> ❓ Skipped this one as I didn't quite understand the aim of the existing private fields support - they seem to be removed completely?

The current state of private class fields isn't great right now. Theoretically, they're "passed through" so that they're supported as long as the runtime supports them, but the class fields transform removes them, which basically causes any uses to break. I think ideally this issue would be either fixed or documented.

Longer-term (or maybe actually pretty soon), I think Sucrase should have a semver-major release to drop the class fields transform since it's fairly well-supported by node and browsers at this point. At that point, supporting private class fields is just a matter of passing it through to the compiled JS.

So I think we *should* actually port this change, with the goal of just passing through the tokens without crashing. From the parser perspective, it should properly emit the token without breaking the parser state.

> bda75 4 months ago Handle private access chained on an optional chain (#11248)
> ❓ Skipped for the same reason as 74590 above

Like above, I think we should port this change so that it works in the parser, but not worry about the runtime implementation. In the future, when Sucrase removes the class fields and optional chaining transforms, it should be able to parse and pass through the syntax so that the underlying JS runtime can evaluate it.

Rugvip:
@alangpierce Thanks! Makes sense to forward the private field tokens, will have a look. Would that just be a `tt.hash` followed by an identifier? I ran into some trouble with `# myVar` being treated the same as `#myVar`, but is it actually in the scope of sucrase to validate that?

alangpierce:
> Would that just be a tt.hash followed by an identifier?

Yep! A reasonable alternative is a separate `privateName` token type, but when in doubt, it's best to stick with the tokens that Babel uses.

> I ran into some trouble with `# myVar` being treated the same as `#myVar`, but is it actually in the scope of sucrase to validate that?

I'd call it out of scope, yes. I looked at the [Babel source code](https://github.com/babel/babel/blob/3ba1ece0ce908cd106c6704175fc56431cb1b32b/packages/babel-parser/src/parser/expression.js#L1275) to confirm what it does, and it parses a `#` token followed by a name token potentially with a space in between, and it then checks if there's a space and throws an error if so. Generally if Babel has specific error-checking code like that, Sucrase leaves it out, which means that sucrase will emit the space and the next downstream tool (which may be webpack or may be the JS engine itself) will give a hopefully-helpful error about the space. A *lot* of complexity in the Babel parser is these validation checks that would have been caught downstream anyway, so I decided that Sucrase should skip all checks like that, since Sucrase's parser tries to be as simple and fast as possible. It would be harder if the space *was* valid syntax and parsed to something different; in that case Sucrase would need to implement it.

Rugvip:
@alangpierce Updated to include support for `#x in obj` and `this?.#x` as discussed

src/parser/tokenizer/index.ts
```diff
@@ -218,6 +218,12 @@ export function lookaheadTypeAndKeyword(): TypeAndKeyword {
   return new TypeAndKeyword(type, contextualKeyword);
 }
 
+// Returns true if the next character is a valid start for an identifier
+export function nextIsIdentifier(): boolean {
```
Rugvip, line 222: Opted for this instead of accessing `input` in the expression traverser, which is essentially what babel does. Should be simple to switch around if you prefer something different.

Rugvip: Never mind, ended up syncing more from babel-parser, which moves over to accessing `input` in the traverser

Rugvip:
@alangpierce Added a couple more commits, should hopefully now have all that we need in the parser to support TS4.

src/parser/plugins/typescript.ts
```diff
@@ -367,6 +367,7 @@ function tsParseMappedType(): void {
 function tsParseTupleType(): void {
   expect(tt.bracketL);
   while (!eat(tt.bracketR) && !state.error) {
+    // Do not validate presence of either none or only labeled elements
```
alangpierce, line 370: This comment doesn't seem super-necessary; across the parser code, all validation is skipped.

alangpierce (approved):
👏 really great work, thank you! I have a few nitpicks, but none particularly important, so I'll merge as-is, then do a release.

>02c8f 3 months ago fix: add optional: false to chained optional call expression (#11814)
>❓ Sets node.optional, assuming this can be skipped since optional chaining is handled differently in Sucrase?

Yep, that sounds right to me! It was just an issue of AST format (`optional` missing vs explicitly false).

src/parser/plugins/typescript.ts
Rugvip: Put it here because it replaces a large chunk of code in babel, but agree that it's pretty useless outside the context of porting babel changes

Issue #550, opened 2020-08-21
Support TypeScript 4.0 and logical assignment operators


andreubotella:
TypeScript 4.0 was released yesterday, and it incorporates a number of additions, some of which Sucrase does not support:

## [Labeled tuple elements](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements)

```ts
type Range = [start: number, end: number];

type Foo = [first: number, second?: string, ...rest: any[]];

function foo(x: [first: string, second: number]) {
    // ...

    // note: we didn't need to name these 'first' and 'second'
    const [a, b] = x;
    a
//  ^ = const a: string
    b
//  ^ = const b: number
}
```
```
SyntaxError: Unexpected token (1:20)
```

## [`unknown` on `catch` clause bindings](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#unknown-on-catch-clause-bindings)

```ts
try {
  // ...
} catch (e: unknown) {
  if (typeof e === "string") {
    // We've narrowed 'e' down to the type 'string'.
    console.log(e.toUpperCase());
  }
}
```
```
SyntaxError: Unexpected token, expected ")" (3:11)
```

## [Logical assignment operators](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#short-circuiting-assignment-operators)

This is, of course, not specific to TypeScript, but [a TC39 proposal](https://github.com/tc39/proposal-logical-assignment) that became Stage 4 in July and is due for inclusion in ECMAScript 2021.

```ts
// These must not compile to the same code!
// (If obj.prop has a setter, it's only called in the logical assignment
// versions if the logical operation doesn't short-circuit.)

obj.prop = obj.prop && foo();
obj.prop &&= foo();

obj.prop = obj.prop || foo();
obj.prop ||= foo();

obj.prop = obj.prop ?? foo();
obj.prop ??= foo();
```

Sucrase outputs:
```js
 function _nullishCoalesce(lhs, rhsFn) { if (lhs != null) { return lhs; } else { return rhsFn(); } }

obj.prop = obj.prop && foo();
obj.prop &&= foo();

obj.prop = obj.prop || foo();
obj.prop ||= foo();

obj.prop = _nullishCoalesce(obj.prop, () => ( foo()));
obj.prop ??= foo();
```

`tsc` and Babel compile these operators down to `obj.prop && (obj.prop = foo());`

[...]

PR #557, opened 2020-10-07
Added transform for logical assignment operators

Rugvip:
Implements transforms of `&&=`, `||=`, and `??=`, as discussed in #550.

This is an early draft to see what you think about the direction. It does end up adding more fields to the tokens, but it's also real tricky to get around that. I did end up with some thoughts about how to refactor into a more generic helper prefixing field, but that'd be a separate thing.

Only tested with transform tests so far, haven't actually executed any transformed code, so needs some more testing there x)


src/transformers/LogicalAssignmentTransformer.ts
```diff
@@ -0,0 +1,164 @@
...
+    resultCode: string;
+    tokenIndex: number;
+  };
+  // The result code position at the start of the computed property access
+  start: number;
+}
+
+export default class LogicalAssignmentTransformer extends Transformer {
+  // This stack stores the state needed to transform computed property access
+  private readonly computedAccessStack: Array<ComputedAccess> = [];
```
Rugvip, line 22: Switched to a manual stack instead of relying on the JS execution stack. This seemed to avoid some of the competition going on with other processors, but could switch back if you think the old approach is better.

aleclarson (approved):

Qix-:
@alangpierce Sorry for the ping, but any chance of getting this published? :)

Rugvip:
@Qix- https://github.com/alangpierce/sucrase/issues/550#issuecomment-695955236 has more context on why the transform is the way it is. Also note that the helper transform is avoided when the left side is a plain variable.

Could be discussed if it's worth merging this btw, as we're approaching support for the syntax in all major runtimes.

Rugvip:
Gonna close this since Node.js v16 is the first LTS release with support for this syntax and it is soon the active LTS. At this point it doesn't make sense to add the complexity of this transform to sucrase, with it being better to rather look at which transforms we can remove.

Qix-:
I don't use Sucrase for Node. I use it for the browser. There are recent browsers that still don't support this syntax.

PR #570, opened 2020-12-11
Fix #560: Support for private class fields

### LINKED ISSUE #560: Support for private class fields
It looks like support was added in TypeScript 3.8: https://www.valentinog.com/blog/private/#private-class-fields-in-typescript

It seems not to be supported yet in `sucrase`. I get the error `Usage of undeclared private name` when I try to use `#` in `sucrase`

nifgraup:
Removed # from tokenizer, consider it to be part of identifier.

### DIFF
--- a/benchmark/sample/expression.ts
+++ b/benchmark/sample/expression.ts
@@ -301,7 +301,7 @@ export function baseParseSubscript(
       parseIdentifier();
     }
   } else if (eat(tt.dot)) {
-    parseMaybePrivateName();
+    parseIdentifier();
   } else if (eat(tt.bracketL)) {
     parseExpression();
     expect(tt.bracketR);
@@ -506,11 +506,6 @@ export function parseExprAtom(): boolean {
     default:
       throw unexpected();
   }
-}
-
-function parseMaybePrivateName(): void {
-  eat(tt.hash);
-  parseIdentifier();
 }
 
 function parseFunctionExpression(): void {
@@ -839,7 +834,7 @@ export function parsePropertyName(objectContextId: number): void {
     if (match(tt.num) || match(tt.string)) {
       parseExprAtom();
     } else {
-      parseMaybePrivateName();
+      parseIdentifier();
     }
 
     state.tokens[state.tokens.length - 1].identifierRole = IdentifierRole.ObjectKey;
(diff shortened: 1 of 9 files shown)


alangpierce:
Thanks for the PR, and sorry for the delay in getting to this! Unfortunately, I think this strategy doesn't quite work. I tried out the new output code from the updated test:
```js
class A {
  constructor() {
    A.prototype.__init.call(this);
  }
  __init() {
    this.#prop2 = value;
  }
}
```
And I get the error in Chrome: `Uncaught SyntaxError: Private field '#prop2' must be declared in an enclosing class`. So I think any working solution here will need to either rewrite each usage of the identifier to not start with a hash, or will need to declare it using plain class field syntax.

That gets into the question of what level of private field support Sucrase *should* have. There are two options here:
1. Sucrase could *pass through* private class field syntax, so that as long as it's supported by the browser or the node version, it will run just fine. In the case above, that means outputting the `#prop2 = value;` class-level declaration and keeping usages as `this.#prop2`.
2. Sucrase could *transpile* private class field syntax (like it does for regular class fields). That means it would rewrite the name to `__private__prop2` or something, but I'd worry about edge cases here. As one nuance, a simple `__private__` prefix would result in incorrect overriding, like in this example:
```js
class A {
  #foo = 3;
  getValue() {
    return this.#foo;
  }
}

class B extends A {
  #foo = 4;
}

const b = new B();
// This needs to print 3 because private fields do not override each other.
console.log(b.getValue());
```

Given the subtleties with option 2 and the fact that recent browsers and node support the syntax, I think option 1 makes the most sense, at least as a first step.

From an implementation standpoint, I'm a bit hesitant about removing `#` from parsing since I'd like it to correspond to Babel's parser as much as is reasonable. I think the right approach for option 1 is to modify `getClassInfo.ts` (which scans classes for class fields to rewrite) to skip over private class field declarations and leave them as-is. I think I'll have some time in the next week to get that approach working, but let me know if you'd prefer to work on it.

Longer-term, I think it makes the most sense to completely drop the class fields transform since it's now widely supported by JS engines, but that would be through a future breaking change.

alangpierce:
Closing since #574 solves this an alternate way.

PR #571, opened 2020-12-28
Add options argument to addHook to pass to pirates.addHook

gordonmleigh:
I've added an optional argument to `addHook` to pass through to pirates. You would use it like this:

```typescript
addHook(".js", { transforms: ["imports"] }, { ignoreNodeModules: false });
```

I also removed the `ts-ignore` comment because pirates has types now.

### DIFF
--- a/src/register.ts
+++ b/src/register.ts
@@ -1,9 +1,13 @@
-// @ts-ignore: no types available.
 import * as pirates from "pirates";
 
 import {Options, transform} from "./index";
 
-export function addHook(extension: string, options: Options): void {
+export interface HookOptions {
+  matcher?: (code: string) => boolean;
+  ignoreNodeModules?: boolean;
+}
+
+export function addHook(extension: string, options: Options, hookOptions?: HookOptions): void {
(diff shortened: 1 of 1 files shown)


alangpierce (approved):
Looks great, thank you!

alangpierce:
To manually test, I added a print to Sucrase to print which files it processes, then ran the require hook from a node shell setting `ignoreNodeModules` to `false`, then I required a third-party package and confirmed that it worked.
```
yarn build
node
> r = require('./dist/register.js')
> r.addHook(".js", {transforms: ["imports"]}, {ignoreNodeModules: false})
> require('prettier')
Transforming /Users/alanpierce/code/sucrase/node_modules/prettier/index.js
Transforming /Users/alanpierce/code/sucrase/node_modules/prettier/third-party.js
```
I'd also be happy to accept a PR extending the param to the other wrappers in `register.ts`.

Unfortunately, looks like CI isn't running anymore, so I'll investigate that as a follow-up, but tests pass for me.

PR #573, opened 2020-12-29
Add options argument to register functions to pass to pirates

gordonmleigh:
Further to #571, this PR exposes the new `addHook` options argument to the `register*` functions too.

Would you consider making the register functions available via the index, so they can be imported without a deep package reference (currently `'sucrase/dist/register'`)? I can add to this PR or another. If we do this I'll also update the readme to document these functions.

Thanks for the quick merge on the other PR!

### DIFF
--- a/src/register.ts
+++ b/src/register.ts
@@ -23,39 +23,47 @@ export function addHook(extension: string, options: Options, hookOptions?: HookO
   );
 }
 
-export function registerJS(): void {
-  addHook(".js", {transforms: ["imports", "flow", "jsx"]});
+export function registerJS(hookOptions?: HookOptions): void {
+  addHook(".js", {transforms: ["imports", "flow", "jsx"]}, hookOptions);
 }
 
-export function registerJSX(): void {
-  addHook(".jsx", {transforms: ["imports", "flow", "jsx"]});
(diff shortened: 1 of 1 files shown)


alangpierce:
> Would you consider making the register functions available via the index, so they can be imported without a deep package reference (currently 'sucrase/dist/register')? I can add to this PR or another. If we do this I'll also update the readme to document these functions.

Unfortunately I think this would break Sucrase in the browser since it would mean a regular Sucrase import would pull in `pirates`, which seems to have a hard dependency on a node environment. I tested it with these steps and saw a webpack failure:
```
yarn build
yarn link
cd website
yarn link sucrase
yarn start
...
Module not found: Error: Can't resolve 'module' in '/Users/alanpierce/code/sucrase/node_modules/pirates/lib'
```
(The package named "module" is built into node, and pirates expects it to exist.)

I might be missing an alternate approach, though; happy to hear suggestions. Another high-level option would be to refactor out `@sucrase/register` and `@sucrase/cli` packages, though IMO it's nice for the basic things to be "batteries included".

alangpierce (approved):
Thanks! As I mentioned, I'm unsure about the best way to export these functions in a cleaner way, so I'll merge this step and happy to discuss the other stuff separately.

PR #581, opened 2020-12-31
Support for passing transform options via register

estrattonbailey:
Hey there! I've been meaning to dig into sucrase for a while now and finally got around to it :)

I'm experimenting with integrating it into [presta](https://github.com/sure-thing/presta) for the development server and I realized that the `sucrase/register` hook does not support transform options.

Is this something y'all would be open to? Is there a better way to go about this than what I have here? I'm happy to make adjustments. Also, if you'd rather not open up these options, all good too. Just thought I'd check!

### DIFF
--- a/src/register.ts
+++ b/src/register.ts
@@ -23,47 +23,49 @@ export function addHook(extension: string, options: Options, hookOptions?: HookO
   );
 }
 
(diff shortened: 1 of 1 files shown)


estrattonbailey:
Closing this since it's easy enough to use `addHook`, as explained in #582. Thanks!

PR #604, opened 2021-03-27
feat: return revert function for registers

antfu:
https://github.com/ariporad/pirates#usage

Allowing users to revert the hook registrations.


alangpierce (approved):
Thank you!

Issue #449, opened 2019-05-06
export default anonymous class with class field


xyc:
Thanks for this great project!

I had an issue when exporting an anonymous class with class field, generated class has `null.prototype.__init.call(this);` in the constructor, which results in an error.

Example reproduction:
https://sucrase.io/#compressedCode=H4sIAPiJz1wAAz2LQQ6DIBBF95zi79SNXqDtpjfwBkSG1AShGUZjYri7QLWbyfv%2F%2FZmXb2DBSHoSWA4LGi7cKNqrMWT16gST0zGCdiFv4m%2Ffv0P%2B9uQFhwI0njhSBs4T4rarbYmyskdbA%2FAw8%2Fa6GPiQc%2BE2w191%2BSaVTgWpPpSeAAAA

```js
import React from 'react'
export default class extends React.Component {
  a = {}
  render() {
    return (
      <div>
        hello
      </div>
    )
  }
}
```

joshjung:
I am also having this issue.

forivall:
This appears to be fixed. In the sandbox, I see the output
```js
"use strict";const _jsxFileName = "sample.tsx";Object.defineProperty(exports, "__esModule", {value: true}); function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; } var _class;var _react = require('react'); var _react2 = _interopRequireDefault(_react);
exports. default = (_class = class extends _react2.default.Component {constructor(...args) { super(...args); _class.prototype.__init.call(this); }
  __init() {this.a = {}}
  render() {
    return (
      _react2.default.createElement('div', {__self: this, __source: {fileName: _jsxFileName, lineNumber: 6}}, "hello"

      )
    )
  }
}, _class)
```

PR #614, opened 2021-05-13
Fix transformation of anonymous "export default" classes

matthieugicquel:
This is a fix for https://github.com/alangpierce/sucrase/issues/449.

I can't really tell if this is the right way / a good way to do this, but it's a small change and tests seem to pass :)

### DIFF
--- a/src/transformers/RootTransformer.ts
+++ b/src/transformers/RootTransformer.ts
@@ -195,7 +195,7 @@ export default class RootTransformer {
     // Both static and instance initializers need a class name to use to invoke the initializer, so
     // assign to one if necessary.
     const needsCommaExpression =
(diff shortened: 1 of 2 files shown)


alangpierce (approved):
Thanks for the fix, and sorry for the delay in getting to this. I thought through the details a bit and I agree with this fix.

The `export default class` syntax is tricky because it's technically a class *statement* to the parser but behaves in many ways like a class *expression* (like having the name optional). When a class statement does have a name, it's *unsafe* to turn it into an expression because then it wouldn't declare the name anymore. But it seems like anonymous class statements can be safely converted to expressions since they weren't declaring a name in the first place.

alangpierce:
Just released as sucrase@3.20.1

PR #615, opened 2021-05-14
[Proposal] Add keepUnusedImports option

mizchi:
## Add this feature

```ts
import { transform } from "sucrase";
const build = transform(`
import Foo from "Foo"; // type only
export function handleFoo(foo: Foo) {
  return foo;
}`, { transforms: ["typescript"], keepUnusedImports: true });

console.log(build.code) // keep Foo imports
```

## Why

I tried to use sucrase in svelte preprocess. (I'm creating compiler in browser to svelte preview. sucrase is most lightweight in typescript compilers)

Svelte use script block variables in template.

```js
<script lang="ts">
  import Foo from "./Foo.svelte";
</script>

<Foo/>
```

Current sucrase default will remove this code because `Foo` is not referenced. I need option for svelte compiler.

## Other case: microsoft typescript and svelte

`svelte-preprocess` has custom typescript plugin to restore imports. https://github.com/sveltejs/svelte-preprocess/blob/main/src/transformers/typescript.ts#L35-L54

And svelte recommends `tsconfig.json`'s `importsNotUsedAsValue: "error"` so users handle imports manually.

### DIFF
--- a/src/Options-gen-types.ts
+++ b/src/Options-gen-types.ts
@@ -26,6 +26,7 @@ export const Options = t.iface([], {
   sourceMapOptions: t.opt("SourceMapOptions"),
   filePath: t.opt("string"),
   production: t.opt("boolean"),
+  keepUnusedImports: t.opt("boolean"),
 });
 
 const exportedTypeSuite: t.ITypeSuite = {
--- a/src/Options.ts
+++ b/src/Options.ts
@@ -46,6 +46,12 @@ export interface Options {
    * If specified, omit any development-specific code in the output.
    */
   production?: boolean;
+
+  /**
+   * If specified, keep unused import specifiers on typescript.
+   * This feature is needed by svelte compiler to tell template block from script block.
+   */
+  keepUnusedImports?: boolean;
 }
 
 export function validateOptions(options: Options): void {
--- a/src/transformers/ESMImportTransformer.ts
+++ b/src/transformers/ESMImportTransformer.ts
@@ -20,7 +20,7 @@ import Transformer from "./Transformer";
 export default class ESMImportTransformer extends Transformer {
   private nonTypeIdentifiers: Set<string>;
   private declarationInfo: DeclarationInfo;
-
+  private keepUnusedImports: boolean = false;
   constructor(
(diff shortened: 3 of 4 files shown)


alangpierce:
Thanks for the contribution! This certainly makes sense to me. I've never been a fan of the auto-removal of imported names, and it certainly makes Sucrase a lot more complicated, though I can see why the TS team made that choice originally.

My initial thought was that this is the equivalent of TypeScript's: `"importsNotUsedAsValues": "preserve",`, but I guess the two are different; regardless of configuration, TS will *always* remove type-only (or unused) name bindings from an import statement, and the `importsNotUsedAsValues` setting just controls whether it removes the import statement as a whole if all the imported names are removed. In contrast, this `keepUnusedImports` Sucrase setting will prevent the name bindings from being removed (and therefore also prevent the import statements from being removed).

One concern I have with this approach is that if an imported name is actually just a type (like an `export type` or `export interface` declaration), then the compiled code would be importing that type it even though the export statement is removed by the compiler in the other file. I *thought* that would typically result in a runtime error, though I may be wrong, and I suppose it depends on the module system implementation. Do you know if that would cause a problem with svelte? One possible approach here could be to use a strategy like in https://github.com/alangpierce/sucrase/pull/433, where we remove an imported value if we see that it's used as a type and only as a type, and leave it alone if it's unused. Another strategy is that everyone writing code with this flag enabled can put all their types in an `import type` statement so that it'll get erased.

Another question on my mind is what to do when compiling to CommonJS (the `imports` transform). The CJS approach doesn't emit variable names in the import statement, it just transforms usages, so arguably the flag doesn't apply. Still, since this flag preserves the side effect of the import on the ESM side, I think it would be best to do the same on the CJS side as well. That would be `CJSImportProcessor.pruneTypeOnlyImports` and `CJSImportProcessor.processImportEquals` in the Sucrase code.

alangpierce:
Thanks again! Now that TS has a `verbatimModuleSyntax` option, I implemented an analogous Sucrase option `keepUnusedImports` (#811) using this PR as a starting point, which I'll release soon in 3.33.0.

PR #619, opened 2021-06-05
Fix for CJSImportProcessor not using interop helper for aliased default imports

Rugvip:
Hi! Found something that I think is a bug in the handling of aliased default imports.

For example, given the following code:

```ts
import { default as myFunc } from './myFunc';

myFunc();
```

We want this to be transpiled to something like this (simplified):

```ts
const myFunc = _interopRequireDefault(require('./myFunc'));

myFunc.default();
```

But since it's not treated as a default import, we get this kind of thing instead:

```ts
const myFunc = require('./myFunc');

myFunc.default();
```

Which may cause trouble if the imported module is not an ESM module.

### DIFF
--- a/src/CJSImportProcessor.ts
+++ b/src/CJSImportProcessor.ts
@@ -178,7 +178,7 @@ export default class CJSImportProcessor {
   private preprocessImportAtIndex(index: number): void {
     const defaultNames: Array<string> = [];
     const wildcardNames: Array<string> = [];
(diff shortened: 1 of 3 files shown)


alangpierce (approved):
Good catch, and the fix looks great, thank you! I also checked the Babel and TS output and confirmed that both of them have this behavior as well.

alangpierce:
Just published as 3.18.2

PR #640, opened 2021-07-29
Add a default export to the jest-plugin

### LINKED ISSUE #638: [@sucrase/jest-plugin]: Jest: a transform must export something.
There's a demo repository which reproduces the error: https://github.com/steschwa/jest-sucrase-demo

## Setup

**package.json**
```json
{
  "devDependencies": {
    "@sucrase/jest-plugin": "^2.1.0",
    "jest": "^27.0.6"
  }
}
```

**jest.config.js**
```js
module.exports = {
  transform: {
    ".tsx?$": "@sucrase/jest-plugin"
  }
}
```

## Problem

After writing some tests and executing them with `yarn jest`, Jest exits almost immediately with 

> Jest: a transform must export something.

After digging around in the `@jest/transformer` and `jest-util` package i've found out that they require a `export default` for custom transformers: https://github.com/facebook/jest/blob/bc50e7f360ab1845abbaa0b3ad788caead0d3174/packages/jest-util/src/requireOrImportModule.ts#L21.

The problem with `@sucrase/jest-plugin` is, that the published file (`dist/index.js`) has a `{ __esModule: true }` which Jest reads in https://github.com/facebook/jest/blob/bc50e7f360ab1845abbaa0b3ad788caead0d3174/packages/jest-util/src/interopRequireDefault.ts#L10 and makes it think that there is a default export but obviously there's not.

## Possible Solution

A simple solution would be to change `export function process(...)` in https://github.com/alangpierce/sucrase/blob/main/integrations/jest-plugin/src/index.ts to `export default { process: function process(...) }`.

Not sure if thats the way to go but it is a quick & easy fix for the problem.

Would you like me to open a PR?

vpontis:
When I try to transform my Jest code with `sucrase`, I get the following error from Jest:

```
Jest: a transform must export something
```

It looks like this is because Jest now expects transformers to have a default export (I'm not sure when this changed).

I published a [test package](https://www.npmjs.com/package/@luma-team/sucrase-jest-plugin-test) and checked that this change works on my own codebase.

I don't expect this will break anything because it preserves the current exports.

Closes #638

### DIFF
--- a/integrations/jest-plugin/src/index.ts
+++ b/integrations/jest-plugin/src/index.ts
@@ -19,3 +19,5 @@ export function process(src: string, filename: string): string {
     return src;
   }
 }
+
+export default {process};


alangpierce (approved):
Thanks, and sorry for the delay! Looks like this was due to a breaking change with Jest 27, and I confirmed that Jest 26 still works fine with or without this change.

PR #642, opened 2021-08-08
Handle shadowed globals when they are JSX names

matthieugicquel:
The existing logic to identify shadowed globals was not used when the shadowed name is a JSX name, even though shadowing can happen in this case.

I encountered that while using sucrase so here's the change that makes this case work


alangpierce (approved):
Thank you, looks good to me!

alangpierce:
Just released as sucrase@3.20.1

PR #643, opened 2021-08-09
[jest-plugin] Fix transform regex in README

srmagura:
Previously, the regex would match filenames that simply contained `js` or `ts`, even if it wasn't the file extension. Like `my_js_logo.png`.


alangpierce (approved):
Thanks!

PR #648, opened 2021-09-24
Narrow the token to async when checking default exports

### LINKED ISSUE #543: Expected async keyword in function export
[Playground link](https://sucrase.io/#code=const%20a%20%3D%20%7B%7D%0A%0Aexport%20default%20a%0A%0Afunction%20b()%20%7B%7D)

```ts
const a = {}

export default a

function b() {}
```

benjdlambert:
Needed to use `tt._async` here to make sure that we check if it's an async function in default exports.

Fixes #543


freben:
@alangpierce Gentle ping! This is a really nice unblocker for us to get our ESM support in place. Thank you.

alangpierce:
Looks like npm authentication is down right now, but I'll publish when it's back up.

PR #653, opened 2021-10-10
[Proposal] Add removeTypeComments option

cyco130:
I have a use case where I want to **generate vanilla JavaScript code samples** (say, usage examples of a library) from TypeScript to **avoid maintaining two versions** of what is essentially the same code. With the right options, Sucrase comes very close. But, unfortunately, it won't remove comments attached to TypeScript-only constructs:

```typescript
// I want this comment to stay

// But I want this one gone, it belongs to the interface (empty line is the heuristic)
interface Foo {}
```

I implemented this behind a new option called `removeTypeComments` and I added a test case.

I think this could quickly become a major use case with, for instance, a Docusaurus or MDX plugin. I have already released [a package](https://www.npmjs.com/package/detype) that does something similar (using Babel + Prettier + a few tricks) but this feels much cleaner.

What do you think?

### DIFF
--- a/src/Options-gen-types.ts
+++ b/src/Options-gen-types.ts
@@ -27,6 +27,7 @@ export const Options = t.iface([], {
   filePath: t.opt("string"),
   production: t.opt("boolean"),
   disableESTransforms: t.opt("boolean"),
+  removeTypeComments: t.opt("boolean"),
 });
 
 const exportedTypeSuite: t.ITypeSuite = {
--- a/src/Options.ts
+++ b/src/Options.ts
@@ -51,6 +51,10 @@ export interface Options {
    * separators, etc.
    */
   disableESTransforms?: boolean;
+  /**
+   * Remove comments attached to removed TypeScript constructs.
+   */
+  removeTypeComments?: boolean;
 }
 
 export function validateOptions(options: Options): void {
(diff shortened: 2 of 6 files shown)


alangpierce:
Hi @cyco130, cool use case, and your `detype` package looks cool.

Unfortunately, I think I'd prefer to keep this out of scope of Sucrase. I think of Sucrase as being focused on transforming *code you write* to *code your JS engine can run*, and it's especially focused on use cases where tool performance can have a significant effect on productivity. The focused use case is what makes Sucrase stay manageable to maintain while keeping great performance. (As a less important note, Sucrase also currently has a guarantee that line numbers match before and after, which is helpful to avoid the need for source maps, so I'd be reluctant to break that with a use case like this.)

Of course, you're welcome to fork the project and extend it for this use case.

As for the implementation, this seems like a pretty good way to do it that handles a lot of cases, though I imagine it may need some further work to handle various edge cases. One example that comes to mind is a case like this:
```ts
class A {
  // some comment
  public foo() {
  }
}
```
Even though `public` is removed, the comment should probably stay in that case, if I'm understanding your use case. Making Sucrase smarter in these situations may require careful tweaks to the parser.

I think AST-based tools (like you're already using with `detype`) will probably give the best results when you want the output code to look nice. By the way, have you seen `recast` and `jscodeshift`? They're designed to allow transforming code while keeping the formatting as-is, so they might fit your use case better than `babel` (which loses existing formatting, if I understand right).

Closing this PR for now, but happy to discuss further if you want.

PR #663, opened 2021-11-04
[jest-plugin] Use transformer config options

stagas:
This PR adds the functionality for the **jest plugin** transformer to use **transformer configuration** options defined in jest.config.

In short, this works now:

```
module.exports = {
  ...
  transform: {
    "\\.(js|jsx|ts|tsx)$": [
      "@sucrase/jest-plugin",
      { jsxPragma: "h", jsxFragment: "Fragment" }
    ],
  ...
}
```
## Notes

Tests are not included as I couldn't find where to put them.

Manually I tested both cases of passing and not passing a config. In the latter case Jest simply sends an empty object `{}` so the spread operation must be safe.

### DIFF
--- a/integrations/jest-plugin/README.md
+++ b/integrations/jest-plugin/README.md
@@ -20,6 +20,15 @@ Then change the default transform in jest.config.js file:
   ...
 ```
 
-Currently, the transforms are not configurable; it uses always runs the import
-and jest transforms and uses the file extension to decide whether to run the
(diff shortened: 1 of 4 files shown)


alangpierce:
Sorry for not getting to this originally! This was also implemented very similarly in #749, which was just merged, so I'll close this PR and also mention this PR as well in the release notes.

PR #678, opened 2022-02-01
fix(docs): Fix README typo, markdown linter problems

lobsterkatie:
This PR contains two changes, corresponding to the two commits included below.

- The actual reason for the PR: Fixes a spot in the `README.md` where the `imports` transform was referred to as the `import` transform.

- A "while I'm here might as well": Adds blank lines before lists and code fence language designations to please the markdown linter, for the markdown files at the top level of the repo. (Other things which were making the linter unhappy, like putting an H3 directly under an H1 without an intervening H2, I left alone, because I didn't actually want to change the appearance of anything.)


alangpierce (approved):
Thanks!

Issue #694, opened 2022-04-27
Plugin does not support Jest 28


fdc-viktor-luft:
Error when running `jest@28.0.1` with configuration:

```js
transform: { "\\.(js|jsx|ts|tsx)$": "@sucrase/jest-plugin" },
```

```
ERROR: ● Invalid return value:
  Code transformer's `process` method must return an object containing `code` key 
  with processed string. If `processAsync` method is implemented it must return 
  a Promise resolving to an object containing `code` key with processed string.
  Code Transformation Documentation:
  https://jestjs.io/docs/code-transformation
```

xalisys:
I am facing the exact same issue, any ideas?

fdc-viktor-luft:
Yes. Luckily it is very easy to fix. I'll create a PR.

alangpierce:
@fdc-viktor-luft Thanks for the PR! I just published the fix as @sucrase/jest-plugin 2.2.1

PR #695, opened 2022-04-28
Jest-Plugin: Support v28

fdc-viktor-luft:
This change is also backward compatible to earlier versions of Jest. Tested it also with Jest@27.

Fixes: https://github.com/alangpierce/sucrase/issues/694

### DIFF
--- a/integrations/jest-plugin/src/index.ts
+++ b/integrations/jest-plugin/src/index.ts
@@ -11,7 +11,11 @@ function getTransforms(filename: string): Array<Transform> | null {
   return null;
 }
 
-export function process(src: string, filename: string): string {
+// this is compatible to the one that is required by Jest, using the type from here:
+// https://github.com/mozilla/source-map/blob/0.6.1/source-map.d.ts#L6-L12
+type RawSourceMap = ReturnType<typeof transform>['sourceMap'];
+
+export function process(src: string, filename: string): {code: string; map?: RawSourceMap | string | null} {
   const transforms = getTransforms(filename);
   if (transforms !== null) {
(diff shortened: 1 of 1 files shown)


integrations/jest-plugin/src/index.ts
```diff
@@ -21,10 +25,10 @@ export function process(src: string, filename: string): string {
     });
     const mapBase64 = Buffer.from(JSON.stringify(sourceMap)).toString("base64");
     const suffix = `//# sourceMappingURL=data:application/json;charset=utf-8;base64,${mapBase64}`;
-    return `${code}\n${suffix}`;
+    return {code: `${code}\n${suffix}`, map: sourceMap};
```
alangpierce, line 28: Now that we're returning a source map, I bet we can remove `suffix` and get rid of the previous two lines. Let me try it to confirm; from https://github.com/alangpierce/sucrase/pull/661 , my way of testing originally was to try it out in WebStorm and see if breakpoints work.

alangpierce: Unfortunately, from some testing, it looks like the `sourceMappingURL` suffix is still necessary for breakpoints to work right in WebStorm, so it looks like it makes sense to keep.

alangpierce (approved):
Thank you! I'll push a quick fix for the prettier errors, then merge and cut a release.

alangpierce:
I tested several versions back in Jest and it seemed to work fine at least back to Jest 24, so I'll consider this non-breaking for semver purposes.

fdc-viktor-luft:
Nice. Thank you. Read your message too late. Fixed the prettier issues 🙂

PR #698, opened 2022-06-03
Bugfix allow re-export of existing named members

cpitt:
Add configurable: true to Object.defineProperty to allow overwriting existing named exports.

Consider the following valid code

```typescript
// lib1.ts
export const a = '1';

// lib2.ts
export const a = '2'

// index.ts
export * from 'lib1.ts'
export * from 'lib2.ts'
export { a } from 'lib2.ts'

```
Currently tsc and babel compile this fine but sucrase fails with

```
TypeError: Cannot redefine property: a
        at Function.defineProperty (<anonymous>)
```

### DIFF
--- a/src/HelperManager.ts
+++ b/src/HelperManager.ts
@@ -26,7 +26,7 @@ const HELPERS: {[name: string]: string} = {
   `,
   createNamedExportFrom: `
     function createNamedExportFrom(obj, localName, importedName) {
-      Object.defineProperty(exports, localName, {enumerable: true, get: () => obj[importedName]});
+      Object.defineProperty(exports, localName, {enumerable: true, configurable: true, get: () => obj[importedName]});
     }
   `,
   // Note that TypeScript and Babel do this differently; TypeScript does a simple existence
@@ -41,7 +41,7 @@ const HELPERS: {[name: string]: string} = {
           if (exports.hasOwnProperty(key)) {
             return;
           }
-          Object.defineProperty(exports, key, {enumerable: true, get: () => obj[key]});
+          Object.defineProperty(exports, key, {enumerable: true, configurable: true, get: () => obj[key]});
         });
     }
   `,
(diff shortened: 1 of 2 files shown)


alangpierce (approved):
Thanks! And sorry for the delay.

This does move some cases a little away from true spec compliance by making the property configurable, though I'd be very surprised if that caused trouble. Generally, Sucrase is ok with being more permissive than other transpilers/implementations, though usually that comes up in terms of compile-time errors that Sucrase skips rather than runtime behavior differences.

There's actually a way to write a test for cases like these using `assertMultiFileOutput` in `imports-test.ts`, so I'll add a quick test for this case in a follow-up and publish.

As part of investigating this, I took a look at the behavior with various situations and different tools:

|  | Your example | Just two star exports | Named export at beginning |
| ------------- | ------------- |---|---|
| Sucrase before | Crashes, `Cannot redefine property: a` | Exports 1, not configurable | Exports 2, not configurable |
| Sucrase after | Exports 2, configurable | Exports 1, configurable | Exports 2, configurable |
| Babel | Exports 2, not configurable | Crashes, `Cannot redefine property: a` | Exports 2, not configurable |
| TypeScript | Exports 2, configurable | Exports 1, not configurable | Exports 2, configurable |
| Node.js 18 ESM | Exports 2, not configurable | Crashes, `The requested module './index.js' contains conflicting star exports for name 'a'` | Exports 2, not configurable |

Some notes from the comparison here and some additional testing:
* It looks like the ideal behavior (assuming node implements this correctly) is that _ambiguous_ star re-exports are disallowed, but if an explicit re-export clears up the ambiguity, then that's fine. So your example code is perfectly valid and should export `'2'`.
* Really, *all* exports are supposed to be non-configurable in ESM, whether it's a star re-export, a named re-export, or a direct export. Node gets this correct, of course, but all transpilers use a configurable property for direct exports, and TS uses a configurable property for named re-exports. So there's certainly precedent for configurable properties, and really in the vast majority of cases (i.e. direct exports) they are configurable in all transpilers.

To be clear, here's how this case works in Babel and TS:
* Babel creates a `_exportNames` object that's hoisted near the top of the file and includes the names of all explicitly-exported values. The `export * from` syntax knows to skip these.
* TypeScript injects a line `exports.a = void 0;` before any import code, which sets `a` as a configurable property, and later `defineProperty` calls keep it configurable. Explicit re-exports take precedence over star re-exports, regardless of ordering, because star re-exports don't modify the exports object if the property already exists.

Both of these approaches are certainly doable in Sucrase (and the babel one seems better if I were to pick), though I think the simpler approach in this PR is the right approach for now.

alangpierce:
Just released as sucrase 3.21.1.

PR #724, opened 2022-07-15
Update README benchmark to latest numbers

alangpierce:
The README benchmark numbers were last updated in January 2021, so this PR
updates it to see what's changed since then.

To help understand the underlying reasons for changes in numbers, I updated
each piece incrementally and re-ran the benchmark at each step. For the numbers
shown here, I ran the full benchmark 3 times and took the best time for each
tool.

Here's the original benchmark:
```
            Time            Speed
Sucrase     1.64 seconds    220221 lines per second
swc         2.13 seconds    169502 lines per second
esbuild     3.02 seconds    119738 lines per second
TypeScript  24.18 seconds   14937 lines per second
Babel       27.22 seconds   13270 lines per second
```

Since running the previous benchmark, I switched from a 2016 Intel MacBook Pro
to a 2020 M1 MacBook Air, so I simulated the original benchmark on my new
laptop by using Sucrase 3.17.0, the pinned versions of all other tools, and
Node 14. Results:
```
Sucrase     0.76 seconds    477950 lines per second
swc         1.05 seconds    344215 lines per second
esbuild     1.43 seconds    253143 lines per second
TypeScript  9.28 seconds    38943 lines per second
Babel       11.03 seconds   32758 lines per second
```
Over a 2x improvement for all tools just from the new laptop! Pretty incredible
in my opinion.

Then I upgraded to Node 18, still running the old version of all tools:
```
Sucrase     0.64 seconds    563963 lines per second
swc         1.04 seconds    348526 lines per second
esbuild     1.42 seconds    254111 lines per second
TypeScript  7.99 seconds    45210 lines per second
Babel       10.3 seconds    35076 lines per second
```
About a 10-20% improvement for tools written in JS just from upgrading Node. As
expected, swc and esbuild aren't affected by the node upgrade.

Then I upgraded all tools to latest (and addressed a breaking change in esbuild
usage):
```
Sucrase     0.65 seconds    558226 lines per second
swc         1.24 seconds    290311 lines per second
esbuild     1.47 seconds    245861 lines per second
TypeScript  8.97 seconds    40258 lines per second
Babel       9.02 seconds    40032 lines per second
```
Interestingly, most tools got a little slower, which isn't too surprising given
that bug fixes and scope creep can gradually hurt performance.

Then I updated the usage of Sucrase and swc to account for better configuration
since last time. Most notably, Sucrase now has a `disableESTransforms` option
that should be used in most cases these days (and I'm hoping to make it the
default in a future semver-major release), and that better matches how all of
the other tools are configured here. Results:
```
Sucrase     0.57 seconds    636975 lines per second
swc         1.19 seconds    304526 lines per second
esbuild     1.45 seconds    248692 lines per second
TypeScript  8.98 seconds    40240 lines per second
Babel       9.18 seconds    39366 lines per second
```
It looks like `disableESTransforms` resulted in about a 15% speedup for Sucrase.

As always, there are plenty of caveats with these measurements, which are
mentioned in the docs in the benchmark file, but I think they still are fair.

One part that I didn't update yet is the Jest codebase being transpiled. I don't
anticipate that having a meaningful effect on the benchmark numbers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,14 +34,14 @@ about 360k lines of code total:
 
 ```text
             Time            Speed
-Sucrase     1.64 seconds    220221 lines per second
-swc         2.13 seconds    169502 lines per second
-esbuild     3.02 seconds    119738 lines per second
-TypeScript  24.18 seconds   14937 lines per second
-Babel       27.22 seconds   13270 lines per second
+Sucrase     0.57 seconds    636975 lines per second
+swc         1.19 seconds    304526 lines per second
+esbuild     1.45 seconds    248692 lines per second
+TypeScript  8.98 seconds    40240 lines per second
+Babel       9.18 seconds    39366 lines per second
 ```
 
-Details: Measured on January 2021. Tools run in single-threaded mode without warm-up. See the
+Details: Measured on July 2022. Tools run in single-threaded mode without warm-up. See the
 [benchmark code](https://github.com/alangpierce/sucrase/blob/main/benchmark/benchmark.ts)
 for methodology and caveats.
 
--- a/benchmark/benchmark.ts
+++ b/benchmark/benchmark.ts
@@ -222,6 +222,7 @@ async function benchmarkFiles(benchmarkOptions: BenchmarkOptions): Promise<void>
         transforms: path.endsWith(".ts")
           ? ["imports", "typescript"]
           : ["jsx", "imports", "typescript"],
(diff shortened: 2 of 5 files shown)


alangpierce:
> Measured change: 18.63% faster

For posterity, this is a result of the config change within the benchmark code, which is used both for the readme and for the CI benchmark.

Issue #667, opened 2021-11-25
Jest plugin doesn't support inline snapshots


fdc-viktor-luft:
### Issue
When using `expect(foo).toMatchInlineSnapshot()` in some test suite, jest complains in the moment it tries to inline the snapshot at the correct position of the code with:

```sh
Jest: Couldn't locate all inline snapshots.
```

### Relations
The issue is very likely related to https://github.com/facebook/jest/issues/6744.

### Further investigations
- I have also projects were this works but I'm using the exact same versions of all packages and I can't tell the difference.
- `expect(foo).toMatchSnapshot()` works without any problems (since Jest doesn't mutate the original code)
- ``expect(foo).toMatchInlineSnapshot(`whatever`)`` works until you would tell Jest to update the snapshot

#### Package versions
```json
"@sucrase/jest-plugin": "^2.2.0",
"jest": "^27.3.1",
```
#### Node version
16.13.0

alangpierce:
Hi @fdc-viktor-luft , thanks for flagging, I just tried it out and I agree that it's an issue. It looks like this was probably caused by https://github.com/alangpierce/sucrase/pull/661 , where I changed the Jest plugin to include a source map as part of the output. I just tried downgrading to `@sucrase/jest-plugin@2.1.1` and confirmed that that fixes the issue. (I also had to downgrade Prettier since it looks like the just-released 2.5.0 doesn't work with Jest.)

From my digging so far, it looks like here's what's happening:

For some background, Sucrase is intentionally careful to have each line of output exactly match the same line of input, so e.g. line 53 in the transformed code is line 53 in the source code and stack traces give helpful results even if you haven't set up source maps. Originally, Sucrase didn't include source maps at all because they weren't necessary, especially if you're only care about line numbers (e.g. when setting debugger breakpoints). However, some tools like WebStorm want to see *some* source map for transformed code, or else they don't let you set breakpoints at all (because they assume the line number would be wrong). To help support these cases, Sucrase is able to generate a simple source map where the start of each line in the output file is mapped to the start of that same line in the input file, basically an "identity" source map.

In order for Jest's inline snapshot feature to work, it needs to know where it can write the snapshot. From my digging so far, it looks like it works by catching the exception and using the stack trace information to figure out the line and column where the updated snapshot text should be placed.

[...]

Issue #585, opened 2021-01-09
Add support for new React 17 JSX transform


alangpierce:
React has a published a description of a new JSX transform with a corresponding runtime library available in new React versions: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html . TypeScript (and others) have already implemented the new transform: https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#jsx-factories . Sucrase should implement some support for it, though the details are still an open question.

Unfortunately, it looks like it's unlikely to simplify the JSX transform for Sucrase, and will likely make it harder. For example:
* For the prod transform, we'll now need to decide whether to emit `jsx` or `jsxs` based on whether the children section (later on in the code) has at least two elements explicitly specified. This sort of "lookahead" behavior isn't well-supported by Sucrase, but could probably be implemented via a special case in the parser and a new token type or a new flag on the token object.
* We'll need to specially detect the `key` prop and reorder it to come after the props. Since the key could be an arbitrary expression that may contain newlines, it will be a challenge to correctly preserve line numbers like the rest of Sucrase does. If we simply move the key expression to after the props (preserving newline chars in each), then we'll end up with broken breakpoints if you try to debug within a callback prop. This issue also affected the class fields implementation before I rewrote it in https://github.com/alangpierce/sucrase/pull/313 to keep the code in order. There might be a clever alternative transform that preserves order, like using a comma expression to assign to an intermediate variable, but it certainly seems like a pain.
* The transform now requires introducing an import and finding an unused name for each imported identifier rather than the old approach of relying on an existing imported name, which likely means more interdependency between the JSX transform and imports transform (even in the ESM case).
* The difference between dev and prod modes is now more significant, with a different `jsxDEV` function that will fail if used in production (from my reading). That will likely make Sucrase's zero-configuration goal harder; if we use dev mode by default, it will completely break in production, and if we use prod mode by default, it will miss useful debugging info in dev.
* The changes to the transform don't make it any more React-agnostic, which I was hoping to see at some point so that other libraries could be used without special configuration.
* Simply having the old and new transforms will make Sucrase configuration a bit awkward. A possible approach is to make it opt-in in Sucrase 3, default in Sucrase 4, and drop the old transform in Sucrase 5, but there could be a more aggressive transition.

There are some positives with the new system, though:
* Third-party library configuration will just be one string (`importSource`) rather than two.
[...]

PR #729, opened 2022-07-27
Add ts-node integration as sucrase/ts-node-plugin

### LINKED ISSUE #726: Add ts-node transpiler plugin
ts-node now allows a configurable transpiler and already has built-in support for swc. It would be great for Sucrase to provide plugin as well. I'm hoping to work on this soon, just filing for tracking purposes.

Docs are here: https://typestrong.org/ts-node/docs/transpilers/

For ts-node users, this would be a nice drop-in replacement to speed up the transpile step. I already prototyped a quick plugin and ran it through a stress test of a 1 million line codebase and confirmed that ts-node with Sucrase is almost 2x faster than ts-node with swc for handling the initial transpile/import. I'll make sure to include a reproducible benchmark as part of this work, similar to the benchmark already on the README.

For Sucrase users, this would be a more polished and feature-complete alternative to `sucrase/register` and `sucrase-node`, which are both fairly simple integrations using the `pirates` library. Most significantly, this would make it easy to use Sucrase for ESM in node via ts-node's loader hook. I've been investigating adding an ESM loader hook for Sucrase, but there are certainly some details to work through, and it would be nice to just integrate with a library that already handles it all. In addition to ESM support, ts-node also has a REPL and would allow configuration (e.g. JSX config) via tsconfig.json.

In terms of implementation, it looks like there are some compile modes that would need to be supported, e.g. targeting commonjs but preserving dynamic `import()`, so those are also options that should be added to Sucrase as part of this work.

Some open questions:
* Should the integration be a separate package (living in the `integrations` directory) or part of the core `sucrase` package? It looks like it should be possible to write without taking on any additional dependencies. Making it part of `sucrase` seems ideal as a way to quickly get started.
* Should this ts-node integration *replace* `sucrase/register` and `sucrase-node` (as a future breaking change)? My leaning is to keep Sucrase's integrations as a lighter-weight way of supporting the common case, but it may make sense to down-scope Sucrase to purely a transpile function, since that's where I've mostly focused anyway. There are also plenty of non-TypeScript cases (e.g. Flow or a Babel-like JSX/imports setup without tsconfig) that Sucrase supports.
* Similarly, is it worth adding a built-in Sucrase loader hook when there's already a ts-node integration? My leaning is yes for similar reasons as above. But this may depend on the details of how complex a loader hook turns out to be.

alangpierce:
Fixes #726

This makes it easy to use Sucrase with all of the nice benefits that ts-node
provides, such as an ESM loader, tsconfig discovery, and a REPL.

The implementation is based off of these docs:
https://typestrong.org/ts-node/docs/transpilers/
and the built-in SWC integration:
https://github.com/TypeStrong/ts-node/blob/main/src/transpilers/swc.ts

Some implementation notes:
* Currently, the plugin is written as a CJS module and written in JS rather than
  TS. This it the easiest approach with the current build system, though it may
  be reasonable to extend the build system to compile a `ts-node-plugin-src`
  directory to `ts-node-plugin` or something like that. The future plan is to add
  an `exports` line to `package.json` in a semver-major release, which will make
  this a bit easier to manage.
* The plugin is included as part of the core `sucrase` package rather than as an
  integration to be installed separately. This should make the usage and version
  management a little easier, and feels reasonable because the integration is
  small and has zero dependencies.
* I rewrote the usage section of the README to put installation instructions at
  the top and to suggest ts-node as the recommended way to use Sucrase with Node.
* I added an `integration-test` directory with various cases that this plugin should
  handle, and it may be useful for future node/tooling integration tests as well.

---

This is a draft PR. Remaining work to do:
- [x] Apply JSX config to Sucrase (with tests)
- [x] Handle `allowJS` and transpiling JS Files (with tests)
- [x] Handle `esModuleInterop`
- [x] Handle errors
- [x] Add better docs on implementation details and caveats


ts-node-plugin/index.js
```diff
@@ -0,0 +1,38 @@
+const {transform} = require("../dist");
+
+const ModuleKindCommonJS = 1;
+
+/**
+ * ts-node transpiler plugin
+ */
+function create(createOptions) {
```
alangpierce, line 8: @cspotcode Here's what I have so far. Note that plenty of tsconfig options aren't supported by Sucrase (e.g. target ES level), so my plan is to just silently go with Sucrase defaults.

In terms of high-level plugin interface feedback, the two-level `create`/`transpile` approach doesn't seem particularly important for my use case, but I can certainly imagine it being useful for other cases. Currently I'm just doing all work in `transpile`, though I suppose I could front-load some of it.

ts-node-plugin/index.js
```diff
@@ -0,0 +1,38 @@
...
+        transforms.push("jsx");
+      }
+
+      const {code, sourceMap} = transform(input, {
+        transforms,
+        sourceMapOptions: {compiledFilename: fileName},
+        filePath: fileName,
+        preserveDynamicImport: nodeModuleEmitKind === "nodecjs",
+        injectCreateRequireForImportRequire: nodeModuleEmitKind === "nodeesm",
+        disableESTransforms: true,
```
alangpierce, line 28: The ES transforms in Sucrase are pretty irrelevant these days when targeting node, so I just disable them (rather than try to read `target` from tsconfig), and I'm planning on removing them in an upcoming breaking change.

ts-node-plugin/index.js
```diff
@@ -0,0 +1,38 @@
...
+      }
+      if (fileName.endsWith(".tsx") || fileName.endsWith(".jsx")) {
+        transforms.push("jsx");
+      }
+
+      const {code, sourceMap} = transform(input, {
+        transforms,
+        sourceMapOptions: {compiledFilename: fileName},
+        filePath: fileName,
+        preserveDynamicImport: nodeModuleEmitKind === "nodecjs",
```
alangpierce, line 26: This was added recently in #727

ts-node-plugin/index.js
```diff
@@ -0,0 +1,38 @@
...
+        injectCreateRequireForImportRequire: nodeModuleEmitKind === "nodeesm",
```
alangpierce, line 27: This was added recently in #728

ts-node-plugin/index.js
```diff
@@ -0,0 +1,38 @@
...
+  const {nodeModuleEmitKind} = createOptions;
+  const {module} = createOptions.service.config.options;
+  return {
+    transpile(input, transpileOptions) {
+      const {fileName} = transpileOptions;
+      const transforms = ["typescript"];
+      if (module === ModuleKindCommonJS || nodeModuleEmitKind === "nodecjs") {
+        transforms.push("imports");
+      }
+      if (fileName.endsWith(".tsx") || fileName.endsWith(".jsx")) {
```
alangpierce, line 18: @cspotcode can you confirm that I'm understanding this right?
* The right way to detect if the JSX transform should be enabled is by looking at the file extension.
* The right way to detect if the file is TypeScript or JavaScript is to look at the file extension. (Unclear if there's much value in disabling the TS transform for JS files, so maybe I'll just leave it enabled for the sake of simplicity, which the swc integration seems to do.)
* ts-node and TS do not allow JSX syntax inside .js files. (I believe for non-TS projects it's fairly common to do that.)

ts-node-plugin/index.js
```diff
@@ -0,0 +1,38 @@
...
+        transforms,
+        sourceMapOptions: {compiledFilename: fileName},
+        filePath: fileName,
+        preserveDynamicImport: nodeModuleEmitKind === "nodecjs",
+        injectCreateRequireForImportRequire: nodeModuleEmitKind === "nodeesm",
+        disableESTransforms: true,
+      });
+      return {
+        outputText: code,
+        sourceMapText: JSON.stringify(sourceMap),
```
alangpierce, line 32: @cspotcode FYI I'm generating the `sourceMap` as a JS object and stringifying it here, and I see that you're later parsing it, modifying it, and stringifying it again, so it's a little wasteful, though I imagine it would be very unlikely to have a measurable impact on performance.

(Sucrase source maps are a bit silly anyway because every line is guaranteed to map to that same line number, but I believe WebStorm requires having some source map in order for debugger breakpoints to work.)

ts-node-plugin/index.js
```diff
@@ -0,0 +1,38 @@
+const {transform} = require("../dist");
+
+const ModuleKindCommonJS = 1;
```
alangpierce, line 3: @cspotcode Having to reference TypeScript's internal enum values is probably the most awkward part of the plugin interface, though it seems very understandable in giving the most flexibility in mapping TS options to transpiler options. I'm assuming it's extremely unlikely that TypeScript would change these values?

ts-node-plugin/index.js
```diff
@@ -0,0 +1,38 @@
...
+  const {module} = createOptions.service.config.options;
+  return {
+    transpile(input, transpileOptions) {
+      const {fileName} = transpileOptions;
+      const transforms = ["typescript"];
+      if (module === ModuleKindCommonJS || nodeModuleEmitKind === "nodecjs") {
+        transforms.push("imports");
+      }
+      if (fileName.endsWith(".tsx") || fileName.endsWith(".jsx")) {
+        transforms.push("jsx");
```
alangpierce, line 19: Currently, Sucrase only supports the old JSX transform; see #585. If the tsconfig is configured for the new transform, an open question is whether to crash with a message that it's unsupported, or if it's better to just transpile with the old transform anyway. Anyone using the new transform likely has at least one file without `import React from "react";` (or is using `jsxImportSource`), so it's probably better to just crash here to give a more helpful error message.

ts-node-plugin/index.js
cspotcode: JSX detection:
looks like you are correct: https://github.com/TypeStrong/ts-node/blob/97f9afd046b66a0fe05a7d76e7a32f94b872016f/src/transpilers/swc.ts#L65-L68
However, this might be a case where ts-node should take responsibility and pass this as a flag to transpilers.  `transpileOptions.isJsx` or something like that.

TS vs JS detection:
Embarrassingly, the swc integration doesn't attempt this at the moment, it assumes everything is TS.  Again, perhaps this is a case where ts-node should pass as an `isTs` flag to make things clearer.

Re JSX syntax inside JS files: Pretty sure you are correct, though at a higher level, my philosophy is to support whatever the TS language service supports.  So if projects are using TS's editor tooling, and it is happy to do tooltips, tab complete, etc with JSX inside JS files, perhaps using specific tsconfig flags, then I want to support that.

ts-node-plugin/index.js
cspotcode: I forget, does sucrase map line numbers but not column numbers?  I might be thinking of something else.

I bet we can adjust ts-node's API to avoid the unnecessary parse-and-stringify.  We can allow transpilers the option of returning either a string or an object.

I have wondered why swc and typescript give us the sourcemap as a string, but I assume it's cuz they do some sort of efficient string-concat emit, so it's faster for them not to build an object.  Not sure though.

ts-node-plugin/index.js
cspotcode: That's my understanding, yeah.  FWIW the swc plugin already duplicates some of those enums.  Not sure if it's helpful for ts-node to expose these lookups as part of the plugin API to make things easier on your end?

https://github.com/TypeStrong/ts-node/blob/97f9afd046b66a0fe05a7d76e7a32f94b872016f/src/transpilers/swc.ts#L81-L129

ts-node-plugin/index.js
cspotcode: A related thing I want to get right: ensuring plugins have a good way to raise diagnostics and / or warnings.  ts-node doesn't do many warnings today, but I want to add a good mechanism for it.  So that ts-node and/or transpiler plugins can warn users about potential config issues, version incompatibilities, stuff like that.  And then we can support config flags to suppress warnings either completely, or suppress warnings by diagnostic code.

What this would mean for sucrase: you have the *option* of returning either errors or warnings from either `create()` or `transpile()` and ts-node will handle either throwing or logging, or suppressing based on tsconfig.  I'll try to jot down ideas in https://github.com/TypeStrong/ts-node/issues/1851

cspotcode:
I left responses to everything above.

One other thing not mentioned:
Today ts-node overrides a few of your compiler options to "just work."  So we force-enable sourcemaps.  But we don't override `"target"` even though we already know, based on node version, which is the target you should probably use.  Cuz we have @tsconfig/bases to reference.

As a breaking change, I'd like to be more automatic about overriding compiler options.  Users can already do this explicitly via `"ts-node": {"compilerOptions": {overrides here` but I'm thinking for most people, they'd prefer it to be automatic?

E.g. for a frontend project where tsconfig is setup for browsers, we can override with `"target": "es2022"` and `"module": "NodeNext"`, stuff like that.  And then add a flag to go back to legacy behavior, something like `"automaticCompilerOptions": false` or something.

ts-node-plugin/index.js
cspotcode: `nodeModuleEmitKind` is marked `@internal` in ts-node's API, so I should update to make this a public part of the API.  Will serve as documentation and as a reminder that we cannot make breaking changes.  I have called this out in https://github.com/TypeStrong/ts-node/issues/1851

ts-node-plugin/index.js
alangpierce: Sucrase source maps only map line numbers right now, which is good enough for most cases. It doesn't try to keep column numbers consistent (e.g. by turning type tokens into spaces), and I think that would be impossible for some transforms, but it could be possible on a best-effort basis in the future. Currently the lack of column information in source maps causes an issue with Jest inline snapshots ( https://github.com/alangpierce/sucrase/issues/667 ), so would be nice to improve, but as I mention in that issue, the fix isn't super obvious.

Not sure about other tools, but FWIW it looks like TypeScript makes an object and stringifies it: https://github.com/microsoft/TypeScript/blob/46a12fdb8ac36db7ace0fdf1929e1b25600ccce7/src/compiler/sourcemap.ts#L50 . For swc, my guess is that maybe it's easier to pass a string than an object between Rust and JS.

ts-node-plugin/index.js
```diff
@@ -0,0 +1,38 @@
...
+      }
+
+      const {code, sourceMap} = transform(input, {
```
alangpierce, line 22: Another thing to think about here is how to handle the TS `esModuleInterop` option, which is enabled by default in Sucrase but can be disabled via the Sucrase `enableLegacyTypeScriptModuleInterop` flag. I've been thinking about dropping support in the next breaking change, though now I'm a little worried about that causing issues for older TS projects that are using `ts-node` and want to try out the `sucrase: true` option. For now, I'll map the option correctly (with a test), and I'm leaning toward keeping the option indefinitely.

ts-node-plugin/index.js
```diff
@@ -0,0 +1,38 @@
...
+    transpile(input, transpileOptions) {
+      const {fileName} = transpileOptions;
+      const transforms = ["typescript"];
+      if (module === ModuleKindCommonJS || nodeModuleEmitKind === "nodecjs") {
+        transforms.push("imports");
+      }
+      if (fileName.endsWith(".tsx") || fileName.endsWith(".jsx")) {
+        transforms.push("jsx");
+      }
+
```
alangpierce, line 21: Another thing to think about is what to do about the TS options `importsNotUsedAsValues: preserve` and `preserveValueImports: true`. Currently, Sucrase just implements the default behavior of eliding type-only (or unused) imports, so there may be a difference in execution behavior if Sucrase elides an import due to all names being type-only, but the import was actually important to run due to import side-effects or due to the order of import execution. To me, this feels like a rare enough situation that maybe it's fine to try to run the code anyway, but a stricter approach would be to give an error/warning if this configuration is detected. My current leaning is to ignore those configs for now. For the future, I'm certainly interested in a "don't elide anything" mode in Sucrase, and can maybe hook up the options then.

@cspotcode it looks to me like the swc integration also ignores those settings.

package.json
```diff
@@ -76,6 +77,14 @@
...
     "ts-interface-checker": "^0.1.9"
   },
+  "peerDependencies": {
+    "ts-node": ">=10.9.1"
+  },
+  "peerDependenciesMeta": {
+    "ts-node": {
+      "optional": true
+    }
+  },
```
alangpierce, line 87: @cspotcode I've been trying to wrap my head around the right way to think about peer dependencies and compatibility expectations here and wanted to get your thoughts and any recommendations.

I based these package.json lines off of the peer dependency you have on SWC; I hadn't previously seen `peerDependenciesMeta` and `optional`, but definitely seems like the right approach for this sort of integration.

The `>=` here is certainly convenient for me because Sucrase won't need to be updated at every ts-node semver-major release, but the worry, of course, is that ts-node breaking changes will break the plugin API. I suppose for the Sucrase case specifically, you'll presumably be careful to not break `sucrase: true`, so we can coordinate whatever migration path makes sense as necessary. But thinking more generally, I could see three ways to think about it:
* The `>=` is correct because ts-node can commit to not breaking old plugins, not even in semver-major releases.
* ts-node may have breaking changes to the plugin interface, but the `>=` is fine because it still provides the guarantee to users that their tools will only break while doing a semver-major upgrade to *some* tool (`ts-node` *or* `sucrase`). The main goal is to avoid a *surprise* break while doing a minor upgrade or regenerating a lockfile, and if `ts-node` breaks the plugin interface, users might need to upgrade `sucrase` or wait for a new version.
* The `>=` is incorrect, and it would be better to list the major versions that I know are compatible and try to keep them updated in a timely way.

Sounds like maybe the second bullet is the most reasonable way to think about it.

If I'm following the advice at https://github.com/semver/semver/issues/502#issuecomment-466501843 , any increase to the minimum ts-node version here is semver-major for Sucrase. It certainly makes sense; re-generating your lockfile (or doing a minor upgrade to one tool) should never break your tools. Some consequences to think about in this case:
* If I release right now and then you later add new convenience features like `isTS` or `isJSX` in a new version, I won't be able to rely on them until the next Sucrase semver-major release, since I need to still support ts-node 10.9.1. Other types of changes like optional diagnostics in the response would be fine for me to add, as long as ts-node doesn't break on extraneous fields now (and I see that it doesn't).
* If ts-node releases a breaking change without providing a path for me to maintain support for old and new ts-node versions in the same plugin, that puts me in a hard spot because it would mean that the only way for me to fix compatibility is with a semver-major upgrade (which I'd generally prefer to avoid). One ad hoc way to avoid this problem is to pick a different function name, e.g. the old API is `create` and the new one is `createTranspiler` and I include both for compatibility for a while and then eventually drop the old `create` in a breaking change. From https://github.com/TypeStrong/ts-node/issues/1851, sounds like you also had some ideas around API version to coordinate this.

Another approach to all of this is to have another package `@sucrase/ts-node-plugin` that people install in addition to the other two. That would make major version bumps less of a big deal since it's just the integration with the semver-major change, but I think including it in the core `sucrase` package certainly seems nicest from a DX perspective.

Maybe I'm overthinking these things and there won't be a need for significant breaking changes, just thinking through how it would all work. Like I mentioned, I'm sure we can coordinate since you wouldn't want `sucrase: true` to break anyway.

My impression from the conversation so far is that breaking changes won't actually be necessary and the existing interface makes additive changes pretty easy, but would be good to know if you have a significant breaking change in mind. To be clear, I'd be fine shipping this as-is without the API interface improvements and then picking them up whenever I do a semver-major upgrade in the future, or I'd also be fine waiting for improvements or breaking changes to the interface if that makes the most sense.

ts-node-plugin/index.js
```diff
@@ -0,0 +1,83 @@
...
+    createOptions.service.config.options;
+
+  // A project with the new JSX transform configured likely has at least one
+  // file with JSX where React is not imported, so fail fast and suggest the
+  // old JSX transform as a workaround.
+  if (jsx === JsxEmitReactJSX || jsx === JsxEmitReactJSXDev) {
+    throw new Error(
+      'The JSX modes "react-jsx" and "react-jsxdev" are not yet ' +
+        'supported by Sucrase. Consider using "react" as a workaround.',
+    );
```
alangpierce, line 40: @cspotcode this is the one error that I have in mind right now, and my impression is that it's most helpful to just fail fast because it means a configuration that's unlikely to work with Sucrase. The Sucrase transpiler itself doesn't give nice diagnostics, it just throws `SyntaxError` (or similar) if it hits a problem, though I could wrap in a `try`/`catch` and return some type of failure if you think that's better.

ts-node-plugin/index.js
alangpierce: Yep, I saw what the swc plugin was doing and copied that approach. Exposing named values (rather than just numbers) could be nice, though I could imagine that being a pain to maintain as TS adds more options and more enum values, and it sounds like I'd only be able to use whatever is defined in the minimum ts-node version that I support. So my impression is that the current approach makes the most sense.

package.json
alangpierce: After some more thought and testing, I'm now thinking that I won't add even an optional peer dependency, since it could break npm users where a project has sucrase and ts-node installed independently without plans to use the plugin. I can always add docs as needed or possibly other approaches to ensure compatibility, but for now will keep it simple.

alangpierce:
@cspotcode Thanks for all the feedback earlier. I'm planning to release this as-is, can certainly iterate on interface details in the future.

Issue #748, opened 2022-09-29
Typescript import-equals is transpiled incorrectly


janmeier:
Sample code:
```ts
import { Foo } from 'foo'
import Bar = Foo

console.log(Foo)
console.log(Bar)
```

tsc output [playground](https://www.typescriptlang.org/play?target=1&module=1#code/JYWwDg9gTgLgBAbzgMQhOBfOAzKERwDk2ahAUKJLHAEICGUcAvCmmWQMYQB2AzhABsApgDoBEAOYAKVBACUnHv2FjJU+lAVA)
```js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
var foo_1 = require("foo");
var Bar = foo_1.Foo;
console.log(foo_1.Foo);
console.log(Bar);
```

sucrase output (using `sucrase@3.27.0`)
```js
"use strict";var _foo = require('foo');

const Bar = Foo

console.log(_foo.Foo)
console.log(Bar)
```

### Context
This is in ts code generated by graphql-codegen, so I cannot easily change it. I found a similar esbuild issue, https://github.com/evanw/esbuild/issues/1283

I hope this is within the scope of sucrase :crossed_fingers:

alangpierce:
I took a closer look and it's harder than I expected, but should still be doable. I made a related fix back in https://github.com/alangpierce/sucrase/pull/441 , but it turns out that this `import =` syntax is the one case in TypeScript where you potentially need to traverse an arbitrary-length chain of import statements to see whether an import should be kept or removed.

If I just change `Foo` to be a regular variable access, then it would start incorrectly preserving the import in cases like this:
```ts
import {Foo} from 'type-only-file-that-doesnt-exist-at-runtime';
import Bar = Foo;
const b: Bar = {};
```

The ideal fix is to just implement TypeScript's behavior, though there may be a simpler fix for just this reported case if I make `Foo` a special identifier type that gets transformed by the imports transform but doesn't count as a reference for import elision reasons (which should be strictly an improvement over the current behavior). I'll explore the options a little more to see what makes the most sense here.

janmeier:
Thanks, I appreciate you looking into this! 

Luckily this only occurs if you import a type in codegen and then rename the type. So in our case we can simply change how types are exported for gql codegen to work around the issue

Shabihashuchi:
Nice to  know

Issue #741, opened 2022-08-21
"Must use import to load ES Module" when trying to use ES module in Jest


ColdSpirit0:
Hi! When I trying to test code that imports ES modules, it fails with error:

```
> node --experimental-vm-modules node_modules/jest/bin/jest.js

(node:5112) ExperimentalWarning: VM Modules is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
 FAIL  tests/index.test.ts
  ● Test suite failed to run

    Must use import to load ES Module: D:\Work\js\test_modules\node_modules\is-path-inside\index.js

    > 1 | import isPathInside from "is-path-inside";
        | ^
      2 |
      3 | export function fun1() {
      4 |     return isPathInside("/aaa/bbb", "/aaa")

      at Runtime.requireModule (node_modules/jest-runtime/build/index.js:1011:21)
      at Object.<anonymous> (src/index.ts:1:1)
      at Object.<anonymous> (tests/index.test.ts:1:1)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        0.76 s
Ran all test suites.
```

Any idea how to fix that?
Example project: https://github.com/ColdSpirit0/jest-sucrase-problem

alangpierce:
Hi @ColdSpirit0 , thanks for reporting and for setting up the example project! I took a look and it looks like the Jest integration needs to be updated to read the Jest options and preserve `import` statements if Jest is running in ESM mode. Should be a relatively straightforward fix I think, and I'll leave this issue open for tracking.

alangpierce:
I just released a new version of the Jest plugin as 3.0.0 with a fix for this. When Jest is configured to treat a file as an ES module, it passes that information into the plugin, which invokes Sucrase with the right configuration. Note that you may need to specify `extensionsToTreatAsEsm: ['.ts'],` in the Jest option to get Jest to treat TS files as ESM.

PR #749, opened 2022-10-04
change(integrations/jest-plugin): Support passing transform options t…

janmeier:
…o the plugin

Make the @sucrase/jest-plugin more configurable, by allowing to specify additional options that are passed to `sucrase.transform`. In our case, we needed to set the `jsxRuntime` to automatic.

I added `@jest/transform` as a dependency to get the correct type for `transformOptions`. It is quite a large package, so let me know if you would rather just hand-type it.

### DIFF
--- a/integrations/jest-plugin/src/index.ts
+++ b/integrations/jest-plugin/src/index.ts
@@ -1,4 +1,7 @@
+import type {TransformOptions} from "@jest/transform";
 import {Transform, transform} from "sucrase";
+
+import type {Options} from "../../../src/Options";
 
 function getTransforms(filename: string): Array<Transform> | null {
   if (filename.endsWith(".js") || filename.endsWith(".jsx")) {
@@ -18,10 +21,12 @@ type RawSourceMap = ReturnType<typeof transform>["sourceMap"];
 export function process(
   src: string,
   filename: string,
+  options: TransformOptions<Options>,
 ): {code: string; map?: RawSourceMap | string | null} {
   const transforms = getTransforms(filename);
   if (transforms !== null) {
     const {code, sourceMap} = transform(src, {
+      ...options.transformerConfig,
       transforms,
       sourceMapOptions: {compiledFilename: filename},
       filePath: filename,
--- a/integrations/jest-plugin/src/index.ts
+++ b/integrations/jest-plugin/src/index.ts
@@ -21,13 +21,13 @@ type RawSourceMap = ReturnType<typeof transform>["sourceMap"];
 export function process(
   src: string,
   filename: string,
-  options: TransformOptions<Options>,
+  options: TransformOptions<Partial<Options>>,
 ): {code: string; map?: RawSourceMap | string | null} {
   const transforms = getTransforms(filename);
   if (transforms !== null) {
     const {code, sourceMap} = transform(src, {
-      ...options.transformerConfig,
       transforms,
+      ...options.transformerConfig,
       sourceMapOptions: {compiledFilename: filename},
       filePath: filename,
     });
--- a/integrations/jest-plugin/README.md
+++ b/integrations/jest-plugin/README.md
@@ -28,6 +28,8 @@ You can specify additional transformation options to sucrase by passing an objec
   ...
 ```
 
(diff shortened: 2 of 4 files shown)


integrations/jest-plugin/src/index.ts
```diff
@@ -18,10 +21,12 @@ type RawSourceMap = ReturnType<typeof transform>["sourceMap"];
 export function process(
   src: string,
   filename: string,
+  options: TransformOptions<Options>,
 ): {code: string; map?: RawSourceMap | string | null} {
   const transforms = getTransforms(filename);
   if (transforms !== null) {
     const {code, sourceMap} = transform(src, {
+      ...options.transformerConfig,
```
alangpierce, line 29: Could you put this one line lower so that it's possible to override `transforms`? It's still a little awkward if you need different transforms for different file extensions, but in most cases that's not necessary. That would enable a workaround for https://github.com/alangpierce/sucrase/issues/741 (at least in many cases) by omitting the `imports` transform when targeting ESM.

integrations/jest-plugin/src/index.ts
```diff
@@ -18,10 +21,12 @@ type RawSourceMap = ReturnType<typeof transform>["sourceMap"];
 export function process(
   src: string,
   filename: string,
+  options: TransformOptions<Options>,
 ): {code: string; map?: RawSourceMap | string | null} {
   const transforms = getTransforms(filename);
   if (transforms !== null) {
```
janmeier, line 27: @alangpierce Re your comment below - Should we change this line to `if (transforms !== null && options.transformerConfig.transforms !== null)`? I can't come up with a case where you want to apply sucrase to other file types than those already covered by `getTransforms`, but :man_shrugging:

alangpierce: I think it's best to leave it as `if (transforms !== null)` since .json files should always be skipped, for example. I think `getTransforms` does need to be updated to account for `.mjs`, `.cjs`, `.mts`, and `.cts`, but that can be a separate fix.

janmeier: Agreed

janmeier:
I added a peer dependency on jest 27, since `transformerConfig` is not available before that. It was introduced here https://github.com/facebook/jest/commit/b0bf8023e97355aecbadecf014ae0e2f806fac0d

janmeier:
> Were you able to test this (using `yarn link` or similar)?

Yep, lint succeed and this change also works in our jest setup at Lenus

janmeier:
Thanks for accepting this PR, and reviewing it quickly! 

With this PR + some changes to our graphql typegen to work around https://github.com/alangpierce/sucrase/issues/748 our tests are now green when running with sucrase :tada: (previously we were running ts-jest)

That means I can now run the tests locally without running out of memory, and they are about 25-50% faster as well :man_dancing:

alangpierce:
Just released a new version of the plugin as 3.0.0

janmeier:
> Just released a new version of the plugin as 3.0.0

:tada:

janmeier:
> Glad to hear it's helpful!
> 
> Also, if you're digging into making Jest tests faster, you might want to try out https://github.com/nicolo-ribaudo/jest-light-runner . Sucrase still uses Mocha for the test suite because every time I've tried to switch to Jest I've run into slowdowns (~250ms -> ~5s for all the tests), I think mainly because of the VM isolation overhead. `jest-light-runner` is basically a Jest mode that disables that, so I've been meaning to try it out for Sucrase, but haven't dug into it much yet. But the perf concerns might also just depend on what sorts of tests you're writing.

I did test `jest-light-runner` some time ago, but was blocked by the fact we needed to upgrade jest. We were on v26 before implemnenting sucrase, because of some major performance issues when using jest v27 and above together with ts-jest and tsc.

Now that we are using sucrase, we can upgrade to the latest version of jest, and might test out `jest-light-runner` after that. 

As I see it, we would have to use `ts-node` with the sucrase integration, since `jest-light-runner` expects everything to be ES modules. Or maybe we could still use `@sucrase/jest-plugin`, given the fixes to ESM made in v3?

PR #755, opened 2022-10-07
Revert jest plugin sucrase peer dependency, update sucrase version

alangpierce:
In #753, I overlooked that Sucrase is already installed as a regular dependency, so there's no need to specify it as a peer dependency. The regular dependency is a little less flexible, but makes installation a little easier, so it's a tradeoff, and for now I'll just keep it the way it was. I do need to upgrade the Sucrase version to a more recent one so that the `preserveDynamicImport` is recognized.


alangpierce:
After thinking more, I actually want to switch this to a peer dependency, so I'll close this and put up a PR for that.

PR #759, opened 2022-11-09
Improve source map generation

### LINKED ISSUE #541: Add column data to source map mappings
Are you planning on adding column information to source map `mappings`?

Some context: I'm using sucrase in a rollup build in a chain of rollup plugins. When building the final source map for the given chunk, rollup attempts to trace the segments by stepping through the chain of sourcemaps that were generated by the plugins. The lack of column information in the sucrase source maps causes this tracing to stop and so causes the source map to be incomplete.

forivall:
Fixes #541

Let me know what kind of refactoring will please you! I'd love to get this in to improve source-map debugging. I implemented it kinda weird to minimize theoretical perf impact when sourcemaps arent requested.


forivall:
On my machine, latest fix brings down the perf loss to

> **Measured change:** 2.6% slower (2.74% slower to 0.94% slower)

alangpierce (commented):
Thank you! I think this will also fix #667 . I still need to take a deeper look at the details, but overall this looks like a good approach.

My main hesitations with detailed source maps in the past have been around complexity and performance, so thanks for implementing it in a way that's relatively lightweight in TokenProcessor. In addition to the automated benchmark (which doesn't include source maps), would you mind doing a comparison of performance without source maps vs simple source maps vs detailed source maps? (You could modify the `sucrase.transform` args in `benchmark.ts` and then run `yarn benchmark jest-dev` with different configs.)

> Additional experimentation shows that if I omit the mappings from TokenProcessor's snapshot(), the perf loss is completely gone. Worth it?

I think it would be good to avoid the `.slice()` somehow. As-is, it adds n^2 time for a file needing many snapshots. Two thoughts:
* You could pass a flag into `TokenProcessor` to know if detailed source maps are enabled, and use that flag to disable any sourcemap-related work if it won't be used. Hopefully the JIT would recognize that the flag is always false and there would be very little overhead in that case.
* I *think* just skipping `.slice()` is actually ok, since all use cases re-walk the same tokens after restore and would set any updated values. It's certainly a little fragile, but I think should be fine with an explanation comment. It's also relatively low-stakes.

Also, I'm curious what your use case is for column mappings (as opposed to just line mappings). I know there's #541 and #667, just curious if you had another use case in mind.

src/Options.ts
```diff
@@ -12,6 +12,7 @@ export interface SourceMapOptions {
    * file.
    */
   compiledFilename: string;
+  simple?: boolean;
```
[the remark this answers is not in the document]
forivall: Absolutely, I just wanted to get things working and feedback on naming, etc. before digging into the polish of full documentation 😄

forivall:
> In addition to the automated benchmark (which doesn't include source maps), would you mind doing a comparison of performance without source maps vs simple source maps vs detailed source maps?

Sure! I'll try get to that on Friday

> I _think_ just skipping `.slice()` is actually ok, since all use cases re-walk the same tokens after restore and would set any updated values. It's certainly a little fragile, but I think should be fine with an explanation comment. It's also relatively low-stakes.

Yeah, that's my thinking too, as I noted [in the commit message](https://github.com/forivall/sucrase/commit/a3dc5c30a3f97274559eb5f703e49e046f5b4f94) (which i forgot to push up 😅 ):

> this _shouldnt_ theoretically cause any issues, as any mappings
generated in the snapshot should be overwritten. if needed, we can clear
out the mappings generated after `this.tokenIndex` in `restoreToSnapshot`

---

My main use case is to improve debugging - for example, the chrome debugger allows setting breakpoints on specific expressions in the middle of a statement. Not super high-stakes, but I was in a mood to look at the code to see if i could figure out a solution, and i found that it was relatively straightforward to add the necessary support code in `TokenProcessor`, and then I really wanted to see if it worked, and was pleased that it did!

alangpierce:
No problem on the delay, thanks for running the numbers!

The 30% slowdown does give me some hesitation, but I confirmed with some extra perf tests that there is minimal/no impact when source maps are disabled, so this change seems at least fine to release in an opt-in way.

I want to think a little more about the right approach from a configuration standpoint. I think there are two open questions here:
* **How many source map modes should there be?** The current PR has three modes, "no source maps", "simple/fast source maps", and "full source maps". That gives the most flexibility, but I do also want Sucrase to be as simple as possible from a configuration standpoint, so I'd maybe lean toward getting rid of the simple mode entirely, especially if the perf difference is minor. 30% vs 10% is maybe not quite "minor", so the best approach doesn't seem obvious. My gut feel is that Sucrase is so fast that usually other tools/operations are the dominating performance issue, so most people would prefer the more correct source maps at the expense of some perf. (An even more aggressive approach would be to get rid of the "no source maps" mode and always generate a full source map, though that would require specifying a file path when calling directly.)
* **What is the default mode?** The current PR makes "full source maps" the default mode, but the most defensive approach would be to keep "simple/fast source maps" as default. I think I agree with "full source maps" as the default, but want to call it out as another decision point.

I also want to take a closer look at the implementation and see if there's a way to speed it up and maybe also avoid the extra dependency. (In a perfect world, Sucrase would have zero dependencies, though a source map library feels pretty defensible.) Given that Sucrase still creates fairly predictable source maps and in most cases makes only minor changes to the code, there might be opportunity for optimization. Here's a profile I ran and at least ~half of the source map time is spent in the library, potentially more.

Screen Shot 2022-11-23 at 9 04 57 AM

That said, it's probably also reasonable to defer additional source map performance work and possibly speed it up later.

src/computeSourceMap.ts
```diff
@@ -33,3 +51,60 @@ export default function computeSourceMap(
...
+    }
+    if (code.charCodeAt(i) === charCodes.lineFeed) {
+      lineStart = i + 1;
+    }
+  }
+  return sourceColumns;
+}
+
+function computeDetailedSourceMap(
+  code: string,
```
alangpierce, line 73: Could you call this parameter `generatedCode`? I think that would be more clear and match the `gen-mapping` terminology.

src/computeSourceMap.ts
```diff
@@ -33,3 +51,60 @@ export default function computeSourceMap(
...
+  filePath: string,
+  {compiledFilename}: SourceMapOptions,
+  rawMappings: Array<number | undefined>,
+  sourceColumns: Array<number>,
+): RawSourceMap {
+  const map = new GenMapping({file: compiledFilename});
+  let j = 0;
+  let currentMapping = rawMappings[j];
+  for (; currentMapping === undefined; currentMapping = rawMappings[++j]);
```
alangpierce, line 82: Could you add a bounds check on `j` in the condition here? As I was trying out different implementation options I ran into an infinite loop here if none of the tokens had mappings.

src/computeSourceMap.ts
```diff
@@ -33,3 +51,60 @@ export default function computeSourceMap(
...
+    maybeAddSegment(map, line, 0, filePath, line, 0);
+  }
+  for (let i = 0; i < code.length; i++) {
+    if (i === currentMapping) {
+      const genColumn = currentMapping - lineStart;
+      const sourceColumn = sourceColumns[j];
+      maybeAddSegment(map, line, genColumn, filePath, line, sourceColumn);
+      for (
+        currentMapping = rawMappings[++j];
+        currentMapping === i || currentMapping === undefined;
```
alangpierce, line 95: Could you add a bounds check on `j` here as well?

src/computeSourceMap.ts
```diff
@@ -33,3 +51,60 @@ export default function computeSourceMap(
...
+    }
+    if (code.charCodeAt(i) === charCodes.lineFeed) {
+      line++;
+      lineStart = i + 1;
+      if (currentMapping !== lineStart) {
+        maybeAddSegment(map, line, 0, filePath, line, 0);
+      }
+    }
+  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
alangpierce, line 107: Not a big deal to have this ignore comment, but you may be able to avoid it by adding `ignoreRestSiblings` to the rule config in `.eslintrc.js`. https://eslint.org/docs/latest/rules/no-unused-vars#ignorerestsiblings

src/computeSourceMap.ts
```diff
@@ -33,3 +51,60 @@ export default function computeSourceMap(
...
+      currentMapping = tokens[++j].start;
+    }
+    if (code.charCodeAt(i) === charCodes.lineFeed) {
+      lineStart = i + 1;
+    }
+  }
+  return sourceColumns;
+}
+
+function computeDetailedSourceMap(
```
alangpierce, line 72: Thanks! The logic all looks good to me, and it's nice to have it self-contained here. It might be better at some point to switch to pre-computing token column indices if source maps will always need all of them, but there's certainly a tradeoff, and this approach seems good.

test/source-maps-test.ts
```diff
@@ -1,27 +1,48 @@
...
+        const outDir = path.join(__dirname, "output");
+        fs.mkdirSync(outDir, {recursive: true});
+        result.sourceMap!.sourcesContent = [source];
+        let suffix = "//# sourceMapping";
+        suffix += `URL=test.js.map`;
+        fs.writeFileSync(path.join(outDir, "test.js"), `${result.code}\n${suffix}`);
+        fs.writeFileSync(path.join(outDir, "test.js.map"), JSON.stringify(result.sourceMap));
+      }
+      const {mappings} = result.sourceMap!;
+      assert.match(mappings, /^[^;]+(;[^;]+){3}$/); // 4 lines
```
alangpierce, line 38: Could you actually just assert the entire string? It looks to me like it isn't so long, and I'd think of it like a snapshot test where any changes (e.g. changes to the column logic or a library upgrade) would be spot-checked, even though it's a bit opaque to someone reading the test.

test/source-maps-test.ts
```diff
@@ -1,27 +1,48 @@
...
     assert.deepEqual(result.sourceMap, {
       version: 3,
       sources: ["test.ts"],
       names: [],
-      mappings: "AAAA;AACA;AACA;AACA",
+      mappings: simple ? simpleMappings : result.sourceMap!.mappings,
       file: "test.js",
     });
+    if (!simple) {
+      if (process.env.WRITE_SOURCE_MAPS) {
```
alangpierce, line 28: How have you been using this, out of curiosity? It might be something that could be added to the demo website instead (though certainly that would be out of scope for this PR), or it might be less necessary if you hard-code the expected mappings.

alangpierce (changes requested):
Thanks! A few small things, but overall looks good! I tested and confirmed that this fixed chrome dev tools inline breakpoints and also fixes Jest's `toMatchInlineSnapshot`.

test/source-maps-test.ts
forivall: I've been using that with source map visualization tools - specifically https://evanw.github.io/source-map-visualization/ and https://sokra.github.io/source-map-visualization/ (via https://www.npmjs.com/package/source-map-visualize) - to validate that the mappings were correct.

alangpierce:
Thanks again for the contribution! I was revisiting this to get it through and pushed an update with a few updates:

* Get rid of simple source map mode. For now, the detailed source maps will be
  the only mode. Also inline `computeDetailedSourceMap` now that it's the only
  case.
* Remove `WRITE_SOURCE_MAPS` in test and related details. In https://github.com/alangpierce/sucrase/pull/793 I added source
  map info to the demo website, which is a nicer system for manually verifying
  correctness.
* Update test code to assert the full source map and a JSON representation of
  the mappings.
* Add integration test for Jest inline snapshots.
* Change lint config to ignore unused rest siblings.
* Add bounds checks and change empty `for` loops to `while` (just a style
  preference).
* Add comments in a few more places.
* Remove unnecessary special case when handling hashbang lines.

alangpierce (approved):

PR #804, opened 2023-07-08
fix: handle specifier as default

sxzz:
Add `module.exports = exports.default;` for case:
```ts
export { x as default } from './foo'
export { x as default }
export { }
export type A = string
export { type A as B }
```

### DIFF
--- a/src/transformers/CJSImportTransformer.ts
+++ b/src/transformers/CJSImportTransformer.ts
@@ -292,6 +292,9 @@ export default class CJSImportTransformer extends Transformer {
       }
       this.processExportDefault();
       return true;
+    } else if (this.tokens.matches2(tt._export, tt.braceL)) {
+      this.processExportBindings();
+      return true;
     }
     this.hadNamedExport = true;
     if (
@@ -314,9 +317,6 @@ export default class CJSImportTransformer extends Transformer {
       this.tokens.matches2(tt._export, tt.at)
     ) {
       this.processExportClass();
-      return true;
-    } else if (this.tokens.matches2(tt._export, tt.braceL)) {
-      this.processExportBindings();
       return true;
     } else if (this.tokens.matches2(tt._export, tt.star)) {
       this.processExportStar();
@@ -798,6 +798,7 @@ export default class CJSImportTransformer extends Transformer {
       }
 
       const specifierInfo = getImportExportSpecifierInfo(this.tokens);
+
       while (this.tokens.currentIndex() < specifierInfo.endIndex) {
         this.tokens.removeToken();
       }
@@ -806,6 +807,12 @@ export default class CJSImportTransformer extends Transformer {
         const exportedName = specifierInfo.rightName;
         const newLocalName = this.importProcessor.getIdentifierReplacement(localName);
         exportStatements.push(`exports.${exportedName} = ${newLocalName || localName};`);
+
+        if (exportedName === "default") {
+          this.hadDefaultExport = true;
+        } else {
+          this.hadNamedExport = true;
+        }
       }
 
       if (this.tokens.matches1(tt.braceR)) {
--- a/src/transformers/CJSImportTransformer.ts
+++ b/src/transformers/CJSImportTransformer.ts
@@ -295,32 +295,6 @@ export default class CJSImportTransformer extends Transformer {
     } else if (this.tokens.matches2(tt._export, tt.braceL)) {
       this.processExportBindings();
       return true;
-    }
-    this.hadNamedExport = true;
-    if (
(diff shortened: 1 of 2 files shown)


src/transformers/CJSImportTransformer.ts
```diff
@@ -798,14 +798,23 @@ export default class CJSImportTransformer extends Transformer {
...
+      if (!specifierInfo.isType) {
         const exportedName = specifierInfo.rightName;
-        const newLocalName = this.importProcessor.getIdentifierReplacement(localName);
-        exportStatements.push(`exports.${exportedName} = ${newLocalName || localName};`);
+        if (exportedName === "default") {
+          this.hadDefaultExport = true;
+        } else {
+          this.hadNamedExport = true;
+        }
+        if (!this.shouldElideExportedIdentifier(specifierInfo.leftName)) {
```
alangpierce, line 813: I think the `shouldElideExportedIdentifier` check should actually be kept with the previous `if` statement, so that implicitly-elided exports are skipped as well. Otherwise, the `module.exports = exports.default;` line will be incorrectly added for this code:
```ts
type T = number;
export {T as default};
```

alangpierce (changes requested):
Thank you! Agreed that this fixes the issues in your description and test, but it looks like splitting up the `if` statement was actually incorrect.

alangpierce:
Also, adding a little context here for myself and future readers, since I was needing to remind myself how `add-module-exports` works:

When using `enableLegacyBabel5ModuleInterop`, analogous to `babel-plugin-add-module-exports`, the expected behavior is to add `module.exports = exports.default;` (making the whole CJS module object the default export) when there is a default export but no named exports. My original implementation missed some cases, which are fixed by this PR:
* Type-only exports should *not* count as a named export for these purposes. Fixed by rearranging some cases so `hadNamedExport` isn't set to `true`, and by considering each export individually in `export {...` lines.
* Explicitly exporting the name `default` should count as a default export, not a named export, which is fixed by doing a string match on the exported name.

src/transformers/CJSImportTransformer.ts
sxzz: If we kept it, the unit test will fail, because the `shouldElideExportedIdentifier` returning value of `export { x as default } from './foo'` is `true`.

---

In your case, we can actually export both a value and a type as default at the same time (though with TS errors)

https://www.typescriptlang.org/play?#code/PTAEAEBcGcFoDsD2BjAFgU2QawFA8gJ4AO6oAKqALyjwCuAtgEboBOA3DugB5GIuSgA3uVABDaKAAm6AGajaAGwEBfDpx58B0uYoEBGIA

But the output will remain only one `export default` and no named export, so we can add `module.exports = exports.default`.

alangpierce: > If we kept it, the unit test will fail

Good point, sorry for not trying that out. I think my case above is still valid, though, see [babel example](https://babeljs.io/repl#?browsers=node%206&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=C4TwDgpgBAKlC8UB2BXAtgIwgJwNwCgIAPMAe22CgG84BDAZygBMIAzWlAG2AF9cg&debug=false&forceAllTransforms=false&modules=commonjs&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Cstage-2%2Cstage-3%2Ctypescript&prettier=false&targets=Node-13&version=7.22.8&externalPlugins=babel-plugin-add-module-exports%401.0.4&assumptions=%7B%7D). If the export is erased due to being a type, it shouldn't count when deciding whether or not to emit `module.exports = exports.default`.

This is actually very similar to #806 that I just implemented, but that fix was just for ESM output since CJS happened to already be working. The core issue, I think, is that `export {...}` and `export {...} from 'foo'` are handled by the same code path, but only the first one should erase type-only exported names. This didn't actually cause any correctness issues for CJS in normal export elision, but it does matter for this case of implementing `enableLegacyBabel5ModuleInterop`. Just like in #806, I think the fix is to look ahead to distinguish the two cases.

This PR certainly puts the feature in better shape, so let me merge it now, and I'll do a quick follow-up to refactor the code from #806 to handle the type case here I mentioned. Thanks again!

> we can actually export both a value and a type as default at the same time

Agreed that this is possible, but I think if `default` is *only* a type, then it should be ignored for the purposes of deciding whether to emit `module.exports = exports.default`. That's consistent with Babel's behavior linked above.

alangpierce (approved):

Issue #825, opened 2024-02-04
Empty template spans break sourcemaps


Septh:
I have set up a repro here: https://github.com/Septh/sucrase-broken-sourcemaps.

After cloning and installing, run `node sucrase.js` to compile the TypeScript code; you can then upload the resulting `index.js` file to Evan Wallace's excellent [source-map-visualization](https://evanw.github.io/source-map-visualization/) tool.

Here's what I get:

![image](https://github.com/alangpierce/sucrase/assets/11949623/2f815bd1-97b9-4b96-835a-7be144e1f28c)
[...]

PR #827, opened 2024-02-06
Fix Template strings break sourcemaps #825

Septh:
Simply filter out empty tokens in `computeSourceColumns()`.

Passes all tests.


Septh:
Sorry, the fix's not a good one, it occasionally crashes. :(

Issue #843, opened 2024-09-17
Miscompile of minified code


segevfiner:
The following code:
```js
async function n(){return await{call:t=>{console.log("call",t)}}}async function a(t){return await(await n())?.call(t),!0}a("1");
```

Compiled with: `pnpm sucrase dist -d dist2 --transforms imports`

Results in:
```js
"use strict"; async function _asyncOptionalChain(ops) { let lastAccessLHS = undefined; let value = ops[0]; let i = 1; while (i < ops.length) { const op = ops[i]; const fn = ops[i + 1]; i += 2; if ((op === 'optionalAccess' || op === 'optionalCall') && value == null) { return undefined; } if (op === 'access' || op === 'optionalAccess') { lastAccessLHS = value; value = await fn(value); } else if (op === 'call' || op === 'optionalCall') { value = await fn((...args) => value.call(lastAccessLHS, ...args)); lastAccessLHS = undefined; } } return value; }async function n(){return await{call:t=>{console.log("call",t)}}}async function a(t){return awaitawait _asyncOptionalChain([(await n()), 'optionalAccess', async _ => _.call, 'call', async _2 => _2(t)]),!0}a("1");
//# sourceMappingURL=test2.js.map%
```

Note the invalid `awaitawait` syntax. I have also seen cases of it producing `returnawait` as well. This is a syntax error and fails to run.

Encountered here https://github.com/egoist/tsup/issues/1091, where tsup uses sucrase with only the `imports` transform to compile to ESM using esbuild and then uses sucrase to transform to CJS as a way to emulate tree shaking with esbuild with CJS, which is not supported natively by it.

PR #844, opened 2024-09-26
fix miscompilation with async optional chaining (#843)

easrng:


bearpong:
hey @alangpierce can we publish this? there's a bug when minifiying code that affects popular library like tsup and esbuild 🙏

PR #836, opened 2024-07-18
Replace `glob` with faster alternative

SuperchupuDev:
`glob` is far from ideal considering its [26](https://npmgraph.js.org/?q=glob%4010.4.5) subdependencies and not that fast speed. This PR replaces its use with [`fdir`](https://www.npmjs.com/package/fdir) (the fastest dir crawler out there) and [`picomatch`](https://www.npmjs.com/package/picomatch). They have both a total of zero (0) subdependencies, making this not just a faster approach, but also one that's *way* lighter.

This PR makes `sucrase` go from [41](https://npmgraph.js.org/?q=sucrase%403.35.0) total subdependencies to just [**16**](https://npmgraph.js.org/?q=https%3A%2F%2Fraw.githubusercontent.com%2Falangpierce%2Fsucrase%2F3f4cb35dc475a20aa287552eec36dc52d3eddf56%2Fpackage.json), reducing its total package size from 5.4 MB to about 1.8 MB :O

Note: since `picomatch` supports an array of patterns, the loop that went over the patterns is not necessary anymore and was removed. Also refactored the skipping file types logic into picomatch's built-in `ignore` option

Relevant prior work: https://e18e.dev/guide/cleanup.html, dotenvx/dotenvx#278

<details><summary>Package size report</summary>

([tool source](https://github.com/TheDevMinerTV/package-size-calculator))

```
Package size report
===================

Package info for "sucrase@3.35.0": 5.4 MB
  Released: 2023-12-22 01:34:43.452 +0000 UTC (30w3d ago)
  Downloads last week: 5,870,223 (68.53%)
  Estimated traffic last week: 32 TB
  Subdependencies: 55

Removed dependencies:
  - glob@10.4.5: 3.6 MB (66.30%)
    Downloads last week: 10,260,938 (N/A% from 10.4.5)
    Downloads last week from "sucrase@3.35.0": 5,870,223 (N/A%)
    Traffic last week: N/A
    Traffic from "sucrase@3.35.0": 32 TB (N/A%)
    Subdependencies: 40 (72.72%)

Added dependencies:
  + fdir@6.2.0: 59 kB (1.09%)
    Downloads last week: N/A (N/A% from 6.2.0)
    Estimated traffic last week: N/A
    Subdependencies: 0 (0%)
  + picomatch@4.0.2: 86 kB (1.59%)
    Downloads last week: 1,805,723 (N/A% from 4.0.2)
    Estimated traffic last week: N/A
    Subdependencies: 0 (0%)

[...]

PR #846, opened 2024-12-23
Switch to tinyglobby

benmccann:
This removes more than half of sucrase's transitive dependencies. Fewer dependencies avoid issues like https://github.com/alangpierce/sucrase/pull/822, reduces supply chain risk, and makes the install smaller for consumers.

https://npmgraph.js.org/?q=glob - 26 dependencies
https://npmgraph.js.org/?q=tinyglobby - 2 dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,10 +73,10 @@
   "dependencies": {
     "@jridgewell/gen-mapping": "^0.3.2",
     "commander": "^4.0.0",
-    "glob": "^10.3.10",
     "lines-and-columns": "^1.1.6",
     "mz": "^2.7.0",
     "pirates": "^4.0.1",
+    "tinyglobby": "^0.2.11",
     "ts-interface-checker": "^0.1.9"
   },
   "engines": {
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,6 +1,6 @@
 /* eslint-disable no-console */
 import commander from "commander";
-import {glob} from "glob";
+import {glob} from "tinyglobby";
 import {exists, mkdir, readdir, readFile, stat, writeFile} from "mz/fs";
 import {dirname, join, relative} from "path";
 
@@ -213,7 +213,7 @@ async function runGlob(options: CLIOptions): Promise<Array<FileInfo>> {
   }
   if (include) {
     for (const pattern of include) {
-      const globFiles = await glob(join(absProject, pattern));
+      const globFiles = await glob(join(absProject, pattern), {expandDirectories: false});
       for (const file of globFiles) {
         if (!file.endsWith(".ts") && !file.endsWith(".js")) {
           continue;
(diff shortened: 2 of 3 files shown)


SuperchupuDev:
i've closed #836 since using tinyglobby is a better approach. can you disable the `expandDirectories` option? #836 has some improvements to this library's code to avoid globbing once for each pattern that could be ported here too :eyes:

benmccann:
Done! Thanks @SuperchupuDev 

@alangpierce any chance you could take a look at this one?

benmccann:
@alangpierce just gently checking in on this. Thanks!

mikemaccana:
Would love to see this, still getting issues like

```
  sucrase  >=3.35.0
  Depends on vulnerable versions of glob
```

since `glob@10.4.5` has more command injection vulnerabilities.

benmccann:
In case it helps, there's a blog post (https://e18e.dev/blog/tinyglobby-migration.html) with lots of details about why you might want to adopt it. Here's an excerpt showing lots of other projects that have made the switch:

A whole raft of popular build tools now rely solely on `tinyglobby` for glob functionality:

- Vite
- SWC
- copy-webpack-plugin
- tsup
- tsdown
- unbuild
- nx
- lerna
- and many more...

None of these tools have a deep dependency on other glob libraries - all of them use `tinyglobby` all the way down 🎉

Similarly, many popular frameworks have made the same move:

- React Router
- Preact
- Angular
- SvelteKit
- Astro
- Starlight
- Eleventy

New apps using these frameworks will no longer have deep dependencies on multiple glob libraries.

For others like Nuxt, SolidStart, and TanStack Start, all dependencies have fully switched to `tinyglobby` and those frameworks will only use `tinyglobby` once they update their dependencies. Most remaining frameworks use both `tinyglobby` and `fast-glob`.

Some ecosystems now rely almost entirely on `tinyglobby` — Svelte maintainer [benmccann](https://www.benmccann.com/) has switched over most of the Svelte ecosystem to using `tinyglobby`. Once the pending PR to `typescript-eslint` is merged, you will be able to setup a new SvelteKit project with [every integration it offers](https://svelte.dev/docs/cli/sv-add#Official-add-ons) and your project will exclusively use `tinyglobby`. Many other notable projects such as `pnpm`, `node-gyp`, and `eslint-import-resolver-typescript` have made the switch as well.

alangpierce:
Thanks so much! Sorry for the radio silence. The blog post is quite compelling, and I appreciate that the `engines` constraint in tinyglobby is quite lenient. I'll merge now and cut a semver-patch release ASAP.

I also noticed that `glob` just released [v13](https://github.com/isaacs/node-glob/blob/main/changelog.md) that removes the CLI, which significantly reduces dependencies. Could be a reasonable alternative, though it seems like tinyglobby is the more popular choice now anyway, so I'll go with this PR. I've been meaning to pull the CLI out of Sucrase as well; that's the only reason globs are needed in the first place, and my impression is Sucrase is much more useful as a lightweight library.

